PR #28, opened 2021-07-09
Updated multiplayer character to not block on dynamic networked rigid…

SergeyAMZN:
… bodies

Signed-off-by: pereslav <pereslav@amazon.com>

### DIFF
--- a/Gem/Code/Source/Components/CharacterComponent.cpp
+++ b/Gem/Code/Source/Components/CharacterComponent.cpp
@@ -52,13 +52,16 @@ namespace MultiplayerSample
             if (isKinematic)
             {
                 const PhysX::ActorData* actorData = PhysX::Utils::GetUserData(rigidDynamic);
-                const AZ::EntityId entityId = actorData->GetEntityId();
-
-                if (NetworkRigidBodyRequestBus::FindFirstHandler(entityId) != nullptr)
+                if (actorData)
                 {
-                    // Network rigid bodies are kinematic on the client but dynamic on the server,
-                    // hence filtering treats these actors as dynamic to support client prediction and avoid desyncs
-                    isKinematic = false;
+                    const AZ::EntityId entityId = actorData->GetEntityId();
+
+                    if (NetworkRigidBodyRequestBus::FindFirstHandler(entityId) != nullptr)
+                    {
+                        // Network rigid bodies are kinematic on the client but dynamic on the server,
+                        // hence filtering treats these actors as dynamic to support client prediction and avoid desyncs
+                        isKinematic = false;
+                    }
                 }
             }
 
--- a/Gem/Code/Source/Components/CharacterComponent.cpp
+++ b/Gem/Code/Source/Components/CharacterComponent.cpp
@@ -10,9 +10,73 @@
 #include <AzFramework/Physics/Character.h>
 #include <Multiplayer/Components/NetworkTransformComponent.h>
 #include <Multiplayer/NetworkTime/INetworkTime.h>
+#include <PhysXCharacters/API/CharacterController.h>
+#include <PhysX/PhysXLocks.h>
+#include <PhysX/Utils.h>
+#include <Source/Components/NetworkRigidBodyComponent.h>
 
 namespace MultiplayerSample
 {
+    bool CollisionLayerBasedControllerFilter(const physx::PxController& controllerA, const physx::PxController& controllerB)
+    {
+        PHYSX_SCENE_READ_LOCK(controllerA.getActor()->getScene());
+        physx::PxRigidDynamic* actorA = controllerA.getActor();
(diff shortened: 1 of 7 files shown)


Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -10,9 +10,73 @@
...
+        // non-kinematic dynamic bodies should not impede the movement of the character
+        if (actor->getConcreteType() == physx::PxConcreteType::eRIGID_DYNAMIC)
+        {
+            const physx::PxRigidDynamic* rigidDynamic = static_cast<const physx::PxRigidDynamic*>(actor);
+
+            bool isKinematic = (rigidDynamic->getRigidBodyFlags() & physx::PxRigidBodyFlag::eKINEMATIC);
+            if (isKinematic)
+            {
+                const PhysX::ActorData* actorData = PhysX::Utils::GetUserData(rigidDynamic);
+                const AZ::EntityId entityId = actorData->GetEntityId();
```
AMZN-Olex, line 55: Is actorData guaranteed to be non-null?

AMZN-Olex (approved):

Gem/Code/Source/Components/CharacterComponent.cpp
SergeyAMZN: good call! It won't be null if created via PhysX gem, but technically game code can create rigid dynamics directly with no user data (or incompatible user data), so added a null check here. thanks!

Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -10,9 +10,76 @@
...
+
+            bool isKinematic = (rigidDynamic->getRigidBodyFlags() & physx::PxRigidBodyFlag::eKINEMATIC);
+            if (isKinematic)
+            {
+                const PhysX::ActorData* actorData = PhysX::Utils::GetUserData(rigidDynamic);
+                if (actorData)
+                {
+                    const AZ::EntityId entityId = actorData->GetEntityId();
+
+                    if (NetworkRigidBodyRequestBus::FindFirstHandler(entityId) != nullptr)
```
kberg-amzn, line 59: You also could just try constructing a network entity handle.. if that succeeds then it's a net bound entity, if it doesn't then it's not. Also we should make sure the entity specifically has a net rigid body component on it, and not a net character component. We should also probably make net character incompatible with net rigid body so we can't put both on the same entity.

Gem/Code/Source/Components/CharacterComponent.h
```diff
@@ -19,6 +20,7 @@ namespace MultiplayerSample
 {
     class CharacterComponent
```
kberg-amzn, line 21: Just a heads up that I've renamed this component to NetworkCharacterComponent upon the advice of the UX team. AnimatedHitVolumes is also now NetworkHitVolumes, so aside from the camera and WASD movement all of our multiplayer sample components are prefixed with 'Network'. It should merge cleanly but it's good to let you know this is coming.

Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -41,6 +108,17 @@ namespace MultiplayerSample
...
+        if (m_physicsCharacter)
+        {
+            auto controller = static_cast<PhysX::CharacterController*>(m_physicsCharacter);
+            controller->SetFilterFlags(physx::PxQueryFlag::eSTATIC | physx::PxQueryFlag::eDYNAMIC | physx::PxQueryFlag::ePREFILTER);
+            if (auto callbackManager = controller->GetCallbackManager())
+            {
+                callbackManager->SetControllerFilter(CollisionLayerBasedControllerFilter);
+                callbackManager->SetObjectPreFilter(CollisionLayerBasedObjectPreFilter);
+            }
+        }
```
AMZN-puvvadar, line 120: Would a client proxy have a physics character? If not, this should be fine. If so, this should probably be limited to autonomous and authority.

kberg-amzn (commented):
Looking good! Consider making the changes to the prefilter, since I suspect multiple player-player collisions won't work correctly with this change as written, but otherwise this looks *far* better.. we can now control fall velocity from within ProcessInput for predictive characters which is great!

Gem/Code/Source/Components/CharacterComponent.cpp
SergeyAMZN: The check here is not if it's net bound or not. We can have net bound kinematic entities that will need to block the characters. 
The check here is for the kinematic rigid bodies that are dynamic on the server (i.e. they are not kinematic from the gameplay perspective). 
That's why we're specifically checking for NetworkRigidBodyRequestBus::FindFirstHandler to see if this entity is a dynamic rigid body on the server.
Good point about making components incompatible!

kberg-amzn (approved):
Awesome

PR #29, opened 2021-07-09
Fix nullptr issues and transform state for autonomous

AMZN-puvvadar:


Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
```diff
@@ -70,20 +70,24 @@ namespace MultiplayerSample
         // inputs for your own component always exist
         WasdPlayerMovementComponentNetworkInput* wasdInput = input.FindComponentInput<WasdPlayerMovementComponentNetworkInput>();
 
-        wasdInput->m_forwardAxis = StickAxis(m_forwardWeight - m_backwardWeight);
-        wasdInput->m_strafeAxis = StickAxis(m_leftWeight - m_rightWeight);
+        if (wasdInput)
```
kberg-amzn, line 73: Seriously this was null? definitely should never ever happen...

AMZN-puvvadar: In some shutdown cases yeah, very odd edge case but worth guarding against.

AMZN-puvvadar:
Moving to a new branch due to DCO

PR #32, opened 2021-07-12
Fix character position and nullptr during teardown

AMZN-puvvadar:

### DIFF
--- a/Gem/Code/Source/Components/CharacterComponent.cpp
+++ b/Gem/Code/Source/Components/CharacterComponent.cpp
@@ -119,7 +119,7 @@ namespace MultiplayerSample
             }
         }
 
-        if (!HasController())
+        if (!HasController() || GetController()->IsAutonomous())
         {
             GetNetworkTransformComponent()->TranslationAddEvent(m_translationEventHandler);
         }
@@ -132,7 +132,22 @@ namespace MultiplayerSample
 
     void CharacterComponent::OnTranslationChangedEvent([[maybe_unused]] const AZ::Vector3& translation)
     {
-        OnSyncRewind();
+        if (m_physicsCharacter == nullptr)
+        {
+            return;
+        }
+
+        const AZ::Vector3 currPosition = m_physicsCharacter->GetBasePosition();
+        if (!currPosition.IsClose(translation))
+        {
+            uint32_t frameId = static_cast<uint32_t>(Multiplayer::GetNetworkTime()->GetHostFrameId());
+            m_physicsCharacter->SetFrameId(frameId);
+            if (GetController() && GetController()->IsAutonomous())
+            {
+                m_physicsCharacter->SetBasePosition(translation);
+                GetEntity()->GetTransform()->SetLocalTranslation(translation);
+            }
+        }
     }
 
     void CharacterComponent::OnSyncRewind()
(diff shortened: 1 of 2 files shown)


Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
```diff
@@ -70,20 +70,24 @@ namespace MultiplayerSample
         // inputs for your own component always exist
         WasdPlayerMovementComponentNetworkInput* wasdInput = input.FindComponentInput<WasdPlayerMovementComponentNetworkInput>();
 
-        wasdInput->m_forwardAxis = StickAxis(m_forwardWeight - m_backwardWeight);
-        wasdInput->m_strafeAxis = StickAxis(m_leftWeight - m_rightWeight);
+        if (wasdInput)
```
kberg-amzn, line 73: Is this really null? Seems like something we should be asserting on.

Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -147,7 +162,6 @@ namespace MultiplayerSample
         {
             uint32_t frameId = static_cast<uint32_t>(Multiplayer::GetNetworkTime()->GetHostFrameId());
             m_physicsCharacter->SetFrameId(frameId);
-            //m_physicsCharacter->SetBasePosition(GetNetworkTransformComponent()->GetTranslation());
```
kberg-amzn, line 150: Note that for now this causes a desync with multiple players, but long term it’s critical we get this working for proper rewind and hit detection.

Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -132,7 +132,22 @@ namespace MultiplayerSample
 
     void CharacterComponent::OnTranslationChangedEvent([[maybe_unused]] const AZ::Vector3& translation)
     {
-        OnSyncRewind();
+        if (m_physicsCharacter == nullptr)
```
kberg-amzn, line 135: This is almost just a copy paste of the sync rewind code.. feels a bit messy like this.

Gem/Code/Source/Components/CharacterComponent.cpp
```diff
@@ -119,7 +119,7 @@ namespace MultiplayerSample
             }
         }
 
-        if (!HasController())
+        if (!HasController() || GetController()->IsAutonomous())
```
kberg-amzn, line 122: If we’re autonomous we don’t need to be calling setBasePosition on the character. We do the setBasePosition call as part of process input, since that’s the only place we should ever be writing predictive state for an autonomous.

AMZN-puvvadar: This is actually specifically for applying a correction which is the only time we do need to. It comes through the same code path we'd see for an entity update which happens to be this.

Gem/Code/Source/Components/CharacterComponent.cpp
AMZN-puvvadar: Agreed, SyncRewind seems to only run server side while OnTranslationChangedEvent can be client. This can probably be collapsed with appropriate ownership checks.

Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
AMZN-puvvadar: Fair point, this is likely hiding a component init order issue or something.

Gem/Code/Source/Components/CharacterComponent.cpp
AMZN-puvvadar: This *might* go away with some of changes Sergey is working on. The issue here is the order of ops between competing systems trying to update transform causes the correction to get blown out. This corrects the other systems to what we need so everything stays consistent.

AMZN-puvvadar:
This change is obsoleted by the new optional automatic PhysX update changes.

PR #42, opened 2021-07-27
Update EngineFinder.cmake

Shatur:
`EngineFinder.cmake` was changed in upstream. Current version appends a dictionaries like array values to `CMAKE_MODULE_PATH`. It results in the path like the following:

```
{
  "path" : "D:/Git/o3de"
}/cmake
```

The PR fixes the issue. `EngineFinder.cmake` was taken from the upstream.

### DIFF
--- a/EngineFinder.cmake
+++ b/EngineFinder.cmake
@@ -1,52 +1,70 @@
 # {BEGIN_LICENSE}
 #
-# Copyright (c) Contributors to the Open 3D Engine Project. For complete copyright and license terms please see the LICENSE at the root of this distribution.
-# 
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 #
 #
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm (approved):

EngineFinder.cmake
```diff
@@ -1,12 +1,9 @@
-# {BEGIN_LICENSE}
```
Shatur, line 1: This line appears when I creating a new project. Is this correct?

amznestebanpapp: looks like is part of the template, however, I dont see a need to have that, if you create an issue in o3de I can remove it from the template

amznestebanpapp (approved):

lumberyard-employee-dm:
@Shatur your commit has failed the digital certificate of origin check.
In order to submit to the O3DE repo you are going to need to amend the commit to signoff on it using the `--signoff` to git commit
https://github.com/o3de/o3de-multiplayersample/pull/42/checks?check_run_id=3175711687

Shatur:
@lumberyard-employee-dm, thanks, done!

PR #48, opened 2021-08-04
Add NetworkedHealthComponent and NetworkRandomComponent

AMZN-puvvadar:
First cut of a Networked Health Component. Looking at weapons, there are points (that look to be WIP) where you could grab this component and use a mix of Get/SetHealth to update values. I went with authoritative vs autonomous since that seems to be how weapons are setup.

### DIFF
--- a/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
@@ -8,8 +8,8 @@
     OverrideInclude="Source/Components/NetworkHealthComponent.h"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
-    <NetworkProperty Type="uint8_t" Name="Health" Init="100" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="true" IsPredictable="false" ExposeToEditor="true" ExposeToScript="false" GenerateEventBindings="true" Description="The amount of health" />
+    <NetworkProperty Type="float" Name="Health" Init="100.0f" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="true" IsPredictable="false" ExposeToEditor="true" ExposeToScript="false" GenerateEventBindings="true" Description="The amount of health" />
 
-    <ArchetypeProperty Type="uint8_t" Name="MaxHealth" Init="100" ExposeToEditor="true" Description="Maximum health" />
+    <ArchetypeProperty Type="float" Name="MaxHealth" Init="100.0f" ExposeToEditor="true" Description="Maximum health" />
 
 </Component>
--- a/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
@@ -3,13 +3,17 @@
 <Component
     Name="NetworkHealthComponent" 
     Namespace="MultiplayerSample" 
-    OverrideComponent="true" 
-    OverrideController="false" 
+    OverrideComponent="false" 
+    OverrideController="true" 
     OverrideInclude="Source/Components/NetworkHealthComponent.h"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
     <NetworkProperty Type="float" Name="Health" Init="100.0f" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="true" IsPredictable="false" ExposeToEditor="true" ExposeToScript="false" GenerateEventBindings="true" Description="The amount of health" />
 
     <ArchetypeProperty Type="float" Name="MaxHealth" Init="100.0f" ExposeToEditor="true" Description="Maximum health" />
+
+    <RemoteProcedure Name="SendHealthDelta" InvokeFrom="Server" HandleOn="Authority" IsPublic="true" IsReliable="false" GenerateEventBindings="false" Description="Health changed RPC">
+        <Param Type="float" Name="deltaHealth" />
+    </RemoteProcedure>
 
 </Component>
--- a/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
@@ -4,7 +4,7 @@
     Name="NetworkHealthComponent" 
     Namespace="MultiplayerSample" 
     OverrideComponent="true" 
-    OverrideController="true" 
+    OverrideController="false" 
     OverrideInclude="Source/Components/NetworkHealthComponent.h"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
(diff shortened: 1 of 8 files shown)


AMZN-puvvadar:
Added NetworkRandomComponent while I'm at it. The current flaw with it is attempting to roll multiple rands on a single instance of the component per HostFrame. Due to how Rewindables work, only one value per HostFrame is supported. This could be solved by moving to an array of ints which will be testable once weapons are integrated.

Gem/Code/Source/AutoGen/NetworkHealthComponent.AutoComponent.xml
```diff
@@ -0,0 +1,15 @@
...
+    Name="NetworkHealthComponent" 
+    Namespace="MultiplayerSample" 
+    OverrideComponent="true" 
+    OverrideController="true" 
+    OverrideInclude="Source/Components/NetworkHealthComponent.h"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+    <NetworkProperty Type="float" Name="Health" Init="100.0f" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="true" IsPredictable="false" ExposeToEditor="true" ExposeToScript="false" GenerateEventBindings="true" Description="The amount of health" />
+
+    <ArchetypeProperty Type="float" Name="MaxHealth" Init="100.0f" ExposeToEditor="true" Description="Maximum health" />
```
kberg-amzn, line 13: We probably want an RPC (invoke from server, handle on authority) that applies a delta to the current health value.. that's what the weapon component would likely invoke when applying damage to various entities hit by a blast.

Gem/Code/Source/Components/NetworkHealthComponent.cpp
```diff
@@ -0,0 +1,70 @@
...
+        ;
+    }
+
+    void NetworkHealthComponent::SetHealth(float updatedHealth)
+    {
+        updatedHealth = AZStd::max(0.f, AZStd::min(updatedHealth, GetMaxHealth()));
+        static_cast<NetworkHealthComponentController*>(GetController())->SetHealth(updatedHealth);
+    }
+
+    void NetworkHealthComponent::OnHealthChangedEvent([[maybe_unused]] const float& health)
```
kberg-amzn, line 50: While this is instructive, you could probably leave this out. Instead we could have components like the NetworkCharacterComponent or a NetworkAiComponent (when that exists) have an event handler that hooks health changes and toggles a dead state or something if health drops below zero. Might help make this a bit cleaner, because then a game could use this component without altering your components code.

Gem/Code/Source/Components/NetworkHealthComponent.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+        // Hook for gameplay events such as player death, player revive, showing damage numbers, etc.
+        ;
+    }
+
+    NetworkHealthComponentController::NetworkHealthComponentController(NetworkHealthComponent& parent)
```
kberg-amzn, line 56: Turn off controller override in the autocomponent.xml if there's no specialized controller logic for this component.

Gem/Code/Source/Components/NetworkRandomComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+                unsigned int i;
+            } u;
+            u.i = r;
+            return u.f - 1.0f;
+    }
+
+    void NetworkRandomComponent::OnSeedChangedEvent([[maybe_unused]] const uint64_t& seed)
+    {
+        // Proxy hook to set rng seed
+        static_cast<NetworkRandomComponentController*>(GetController())->SetSeed(seed);
```
kberg-amzn, line 87: Any idea when this would get triggered or what it might be used for? It looks like it's setting the network property to the value we just finished setting and are notifying we've set the network property to. Basically, entityupdate -> update seed property -> fire event -> update seed property?

Gem/Code/Source/Components/NetworkRandomComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+        m_simpleRng.SetSeed(seed);
+        m_seedInitialized = true;
+    }
+
+    NetworkRandomComponentController::NetworkRandomComponentController(NetworkRandomComponent& parent)
```
kberg-amzn, line 92: You want all your GetRandom calls to live here, and not on the component.

Gem/Code/Source/Components/NetworkRandomComponent.h
```diff
@@ -0,0 +1,51 @@
...
+
+        static void Reflect(AZ::ReflectContext* context);
+
+        NetworkRandomComponent();
+
+        void OnInit() override;
+        void OnActivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
+        void OnDeactivate(Multiplayer::EntityIsMigrating entityIsMigrating) override;
+
+        uint64_t GetRandomUint64();
```
kberg-amzn, line 30: Move all of this to the controller. Multiplayer components are read-only by design.

kberg-amzn (changes requested):
Some quick feedback, but overall this is really close and looking quite good!

Gem/Code/Source/Components/NetworkHealthComponent.cpp
AMZN-puvvadar: I could go either way on this. My idea here was to provide more specific handlers here other components could hook so users don't end up with a single bulky health handler. The only adds they'd need here are if they had a scenario we didn't otherwise capture.

Gem/Code/Source/Components/NetworkRandomComponent.cpp
AMZN-puvvadar: The thought here was syncing a server update on the seed to the client as SimpleLcgRandom needs the update at some point. I can dig into this more.

Gem/Code/Source/Components/NetworkRandomComponent.h
AMZN-puvvadar: Okay, good to know.

Gem/Code/Source/Components/NetworkRandomComponent.cpp
kberg-amzn: This would be okay.. changing the seed on the server is a guaranteed desync and the client would take the updated seed from a correction.

Gem/Code/Source/Components/NetworkRandomComponent.cpp
```diff
@@ -0,0 +1,63 @@
...
+                unsigned int i;
+            } u;
+            u.i = r;
+            return u.f - 1.0f;
+    }
+
+    NetworkRandomComponentController::NetworkRandomComponentController(NetworkRandomComponent& parent)
+        : NetworkRandomComponentControllerBase(parent)
+    {
+        if (IsAuthority())
```
kberg-amzn, line 44: I thought @AMZN-Gene changed the naming of these functions to reduce confusion..? Not your change, but this might indicate we missed some? Gene renamed these guys to IsNetRoleAuthority(), IsNetRoleAutonomous(), etc.. specifically to avoid any confusion regarding IfClient() or IfServer() [now IsNetRoleClient() and IsNetRoleServer() respectively].

kberg-amzn (approved):
Nice! Looking really good!

Gem/Code/Source/Components/NetworkRandomComponent.cpp
AMZN-puvvadar: It's on the components only. It's a good point, we should update the controllers as well. In fact:

```
bool MultiplayerController::IsAuthority() const
{
    return GetNetBindComponent() ? GetNetBindComponent()->IsNetEntityRoleAuthority() : false;
}
```

SergeyAMZN (approved):

PR #52, opened 2021-08-10
fix for w4018

amznestebanpapp:
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4018?view=msvc-160

```
Using the token operator to compare signed and unsigned numbers required the compiler to convert the signed value to unsigned
```

### DIFF
--- a/Gem/Code/Source/Components/NetworkAnimationComponent.cpp
+++ b/Gem/Code/Source/Components/NetworkAnimationComponent.cpp
@@ -100,16 +100,16 @@ namespace MultiplayerSample
 
         if (m_velocityParamId == InvalidParamIndex)
         {
-            m_velocityParamId = m_animationGraph->FindParameterIndex(GetVelocityParamName().c_str());
-            m_aimTargetParamId = m_animationGraph->FindParameterIndex(GetAimTargetParamName().c_str());
-            m_crouchParamId = m_animationGraph->FindParameterIndex(GetCrouchParamName().c_str());
-            m_aimingParamId = m_animationGraph->FindParameterIndex(GetAimingParamName().c_str());
-            m_shootParamId = m_animationGraph->FindParameterIndex(GetShootParamName().c_str());
-            m_jumpParamId = m_animationGraph->FindParameterIndex(GetJumpParamName().c_str());
-            m_fallParamId = m_animationGraph->FindParameterIndex(GetFallParamName().c_str());
-            m_landParamId = m_animationGraph->FindParameterIndex(GetLandParamName().c_str());
-            m_hitParamId = m_animationGraph->FindParameterIndex(GetHitParamName().c_str());
-            m_deathParamId = m_animationGraph->FindParameterIndex(GetDeathParamName().c_str());
+            m_velocityParamId = static_cast<uint32_t>(m_animationGraph->FindParameterIndex(GetVelocityParamName().c_str()));
+            m_aimTargetParamId = static_cast<uint32_t>(m_animationGraph->FindParameterIndex(GetAimTargetParamName().c_str()));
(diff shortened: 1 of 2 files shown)


Gem/Code/Source/Components/NetworkAnimationComponent.cpp
```diff
@@ -100,16 +100,16 @@ namespace MultiplayerSample
...
-            m_aimTargetParamId = m_animationGraph->FindParameterIndex(GetAimTargetParamName().c_str());
-            m_crouchParamId = m_animationGraph->FindParameterIndex(GetCrouchParamName().c_str());
-            m_aimingParamId = m_animationGraph->FindParameterIndex(GetAimingParamName().c_str());
-            m_shootParamId = m_animationGraph->FindParameterIndex(GetShootParamName().c_str());
-            m_jumpParamId = m_animationGraph->FindParameterIndex(GetJumpParamName().c_str());
-            m_fallParamId = m_animationGraph->FindParameterIndex(GetFallParamName().c_str());
-            m_landParamId = m_animationGraph->FindParameterIndex(GetLandParamName().c_str());
-            m_hitParamId = m_animationGraph->FindParameterIndex(GetHitParamName().c_str());
-            m_deathParamId = m_animationGraph->FindParameterIndex(GetDeathParamName().c_str());
+            m_velocityParamId = static_cast<uint32_t>(m_animationGraph->FindParameterIndex(GetVelocityParamName().c_str()));
```
jeremyong-az, line 103: Do we prefer `az_numeric_cast`?

amznestebanpapp: in this case I can switch it, but was leaving it to the code owners to do what is best... e.g. a better solution could be to change those Ids to the proper type

jeremyong-az (approved):

Gem/Code/Source/Components/NetworkAnimationComponent.cpp
kberg-amzn: you can leave this as is, though my preference would be to update these types in the class definition to match the new emfx return value type (AZStd::size_t?)

kberg-amzn: actually yeah, since the invalid return type is now (uint64_t)(-1).. we should be using the EMFX types here to avoid bugs.

amznestebanpapp: changed

kberg-amzn (approved):

lumberyard-employee-dm (approved):

Gem/Code/Source/Components/NetworkAnimationComponent.h
```diff
@@ -71,15 +71,15 @@ namespace MultiplayerSample
...
+        size_t m_velocityParamId = InvalidParamIndex;
+        size_t m_aimTargetParamId = InvalidParamIndex;
+        size_t m_crouchParamId = InvalidParamIndex;
+        size_t m_aimingParamId = InvalidParamIndex;
+        size_t m_shootParamId = InvalidParamIndex;
+        size_t m_jumpParamId = InvalidParamIndex;
+        size_t m_fallParamId = InvalidParamIndex;
+        size_t m_landParamId = InvalidParamIndex;
+        size_t m_hitParamId = InvalidParamIndex;
+        size_t m_deathParamId = InvalidParamIndex;
```
lumberyard-employee-dm, line 83: One easy to solve updating this ever again is with `decltype(InvalidParamIndex)`

lumberyard-employee-dm (approved):

PR #59, opened 2021-08-30
Multiplayer weapons

kberg-amzn:
Weapons are now reasonably functional

### DIFF
--- a/Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
+++ b/Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
@@ -27,8 +27,16 @@ namespace MultiplayerSample
         NetworkWeaponsComponentBase::Reflect(context);
     }
 
+    NetworkWeaponsComponent::NetworkWeaponsComponent()
+        : NetworkWeaponsComponentBase()
+        , m_activationCountHandler([this](int32_t index, uint8_t value) { OnUpdateActivationCounts(index, value); })
+    {
+        ;
+    }
+
     void NetworkWeaponsComponent::OnInit()
     {
+        AZStd::uninitialized_fill_n(m_fireBoneJointIds.data(), MaxWeaponsPerComponent, InvalidBoneId);
     }
 
     void NetworkWeaponsComponent::OnActivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
@@ -46,6 +54,11 @@ namespace MultiplayerSample
             m_weapons[weaponIndex] = AZStd::move(CreateWeapon(constructParams));
         }
 
+        if (IsNetEntityRoleClient())
+        {
+            ActivationCountsAddEvent(m_activationCountHandler);
+        }
+
         if (m_debugDraw == nullptr)
         {
             m_debugDraw = DebugDraw::DebugDrawRequestBus::FindFirstHandler();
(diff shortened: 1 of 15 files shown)


Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -68,31 +116,68 @@ namespace MultiplayerSample
...
+                    10.0f
+                );
+            }
+
+            AZLOG
+            (
+                NET_Weapons,
+                "Predicted hit on entity %u at position %f x %f x %f",
+                hitEntity.m_hitNetEntityId,
+                hitEntity.m_hitPosition.GetX(),
```
AMZN-Olex, line 173: Fyi, you can also use AZ::Vector3().ToString().c_str() here with %s.

AMZN-Olex (approved):

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -28,8 +28,16 @@ namespace MultiplayerSample
         NetworkWeaponsComponentBase::Reflect(context);
     }
 
+    NetworkWeaponsComponent::NetworkWeaponsComponent()
+        : NetworkWeaponsComponentBase()
+        , m_activationCountHandler([this](int32_t index, uint8_t value) { OnUpdateActivationCounts(index, value); })
+    {
+        ;
```
jeremyong-az, line 35: typo?

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -55,10 +68,45 @@ namespace MultiplayerSample
 
     void NetworkWeaponsComponent::OnDeactivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
     {
+        ;
```
jeremyong-az, line 71: Typo?

Gem/Code/Source/Weapons/WeaponTypes.cpp
```diff
@@ -363,7 +363,7 @@ namespace MultiplayerSample
 
     bool FireParams::operator!=(const FireParams& rhs) const
     {
-        return m_targetPosition.IsClose(rhs.m_targetPosition)
+        return !m_targetPosition.IsClose(rhs.m_targetPosition)
```
jeremyong-az, line 366: Haha nice catch

jeremyong-az (approved):
Note: my review is of questionable value but it was good reading through to see the general code topology of the system

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -55,10 +68,45 @@ namespace MultiplayerSample
...
+        if (m_fireBoneJointIds[weaponIndexInt] == InvalidBoneId)
+        {
+            const char* fireBoneName = GetFireBoneNames(weaponIndexInt).c_str();
+            m_fireBoneJointIds[weaponIndexInt] = GetNetworkAnimationComponent()->GetBoneIdByName(fireBoneName);
+        }
+
+        AZ::Transform fireBoneTransform;
+        if (!GetNetworkAnimationComponent()->GetJointTransformById(m_fireBoneJointIds[weaponIndexInt], fireBoneTransform))
+        {
+            AZLOG_WARN("Failed to get transform for fire bone %s, joint Id %u", GetFireBoneNames(weaponIndexInt).c_str(), m_fireBoneJointIds[weaponIndexInt]);
```
AMZN-puvvadar, line 100: Is the default just identity here? Is there a more meaningful alternative value here? Maybe just object transform?

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -68,31 +116,68 @@ namespace MultiplayerSample
 
     void NetworkWeaponsComponent::OnWeaponActivate([[maybe_unused]] const WeaponActivationInfo& activationInfo)
     {
+        // If we're replaying inputs then early out
```
AMZN-puvvadar, line 119: Is this because it's debug draw, because we expect replay to manage projectile life cycle or both?

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
kberg-amzn: I'll have to remember that, useful.. thanks!

Gem/Code/Source/Weapons/WeaponTypes.cpp
kberg-amzn: yeah, that actually caused target positions to not replicate correctly to simulated clients :(

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
kberg-amzn: It essentially means the data is bad, but it's not unrecoverable. The simulation will behave incorrectly but it will not crash.

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
kberg-amzn: These callbacks will invoke during corrections if we replay inputs where weapon activations occurred. This means you shoot a weapon, get cool muzzle flash effects and such, immediately get a correction and your weapon would retrigger all those muzzle flash effects. Very bad visual bug. This guard here just ensures that any effects, particles, sounds, debug draw, whatever presentation logic we want to attach here gets correctly discarded during a correction.

PR #60, opened 2021-09-02
Various fixes and improvements to multiplayer weapons

kberg-amzn:
Weapons are now reasonably functional


AMZN-puvvadar (approved):

Gem/Code/Source/Components/NetworkCharacterComponent.cpp
```diff
@@ -182,6 +183,16 @@ namespace MultiplayerSample
 
     AZ::Vector3 NetworkCharacterComponentController::TryMoveWithVelocity(const AZ::Vector3& velocity, float deltaTime)
     {
+        // Ensure any entities that we might interact with are properly synchronized to their rewind state
```
AMZN-Gene, line 186: this chunk of code is showing up in a lot of places, makes me think we can make it a util function. :P

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -16,6 +17,9 @@
 namespace MultiplayerSample
 {
     AZ_CVAR(bool, cl_WeaponsDrawDebug, true, nullptr, AZ::ConsoleFunctorFlags::Null, "If enabled, weapons will debug draw various important events");
+    AZ_CVAR(float, cl_WeaponsDrawDebugSize, 0.25f, nullptr, AZ::ConsoleFunctorFlags::Null, "The size of sphere to debug draw during weapon events");
+    AZ_CVAR(float, cl_WeaponsDrawDebugDurationSec, 10.0f, nullptr, AZ::ConsoleFunctorFlags::Null, "The number of seconds to display debug draw data");
+    AZ_CVAR(float, sv_ImpulseScalar, 750.0f, nullptr, AZ::ConsoleFunctorFlags::Null, "A fudge factor for imparting impulses on rigid bodies due to weapon hits");
```
AMZN-Gene, line 22: sv_ImpulseScalar = sv_WeaponsImpulseScalar (just for consistency?)  I sort of wish all our cl_ and sv_ was prefixed with a net_ or something, this way i could search net_ for all the useful network cvars or net_sv/net_cl to narrow it down

AMZN-Gene (approved):

Gem/Code/Source/Components/NetworkCharacterComponent.cpp
kberg-amzn: Yup, it's pretty important. Currently it shows up in two places, but it will show up anywhere where we need to sync rewound entity state with an external library, like physx or recast for pathfinding, etc.. All it does is compute the AABB of the area you're about to interact with, and then call into the network time system to find any potential overlapping entities to tell them to sync.

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
kberg-amzn: Good call, done.

Gem/Code/Source/Components/NetworkCharacterComponent.cpp
AMZN-Gene: 👍

PR #62, opened 2021-09-07
Implement initial performance test harness

jeremyong-az:
## StressTestComponent

The StressTestComponent is intended to provide an ImGUI interface for
interface with any stress testing facilities/functions provided in the
future. It is provided as a network component to support RPCs and any
properties we wish to replicate that are not specific to other game
entities. Currently, it provides a single RPC to spawn N animated
entities.

There is an unknown bug preventing this component from being
instantiated on the client, so at this time, AI entities may only be
spawned on the server.

image

## NetworkAiComponent

The NetworkAiComponent is attached to the player entity and disabled by
default. When enabled, the WasdMovementComponent and
NetworkWeaponsComponent on the player entity are no longer driven by
inputs. Instead, they tick and poll for the input state from the
NetworkAiComponent.

Currently, the AI component is a chaos monkey, randomizing the current
movement direction, action, and weapon state at random intervals. A
future change could set the seed and decouple the action from the
deltaTime to create reproducible tests.

image

## Miscellaneous changes

- Fixed the material on the animating character
- Commits the project physmaterial
- Saving the level prefab with the most recent editor changes the material formats saved with LOD metadata


Gem/Code/Source/Components/NetworkAiComponent.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    void NetworkAiComponent::TickMovement(WasdPlayerMovementComponentController& movementController, float deltaTime)
+    {
+        // TODO: Execute this tick only if this component is owned by this endpoint (currently ticks on server only)
+        m_remainingTimeMs -= deltaTime * 1000;
+
+        if (m_remainingTimeMs <= 0)
+        {
+            // Determine a new directive after 500 to 9500 ms
+            m_remainingTimeMs = m_lcg.GetRandomFloat() * 9500.f + 500.f;
```
kberg-amzn, line 51: those numbers would make good cvars.. we can hook cvars up to imgui sliders in menus as well

Gem/Code/Source/Components/NetworkWeaponsComponent.h
```diff
@@ -68,6 +70,7 @@ namespace MultiplayerSample
     class NetworkWeaponsComponentController
         : public NetworkWeaponsComponentControllerBase
         , private StartingPointInput::InputEventNotificationBus::MultiHandler
+        , private AZ::TickBus::Handler
```
kberg-amzn, line 73: It might be nicer to use scheduled events for these sorts of polling operations?  AZ::ScheduledEvent.  This way if you spawn a few million of these critters there will be some amount of batching and prioritization so we can maintain a consistent tick rate rather than forcibly linearly visiting every registered component.

Gem/Code/Source/Components/NetworkAiComponent.cpp
jeremyong-az: I actually intended on hooking them up to ImGUI sliders but could not get the StressTestComponent to replicate to the client (which is where I would host that data). I can change them to cvars at the minimum before checkin.

Gem/Code/Source/Components/NetworkWeaponsComponent.h
jeremyong-az: Haven't heard of that till now, will take a look thanks!

Gem/Code/Source/Components/StressTestComponent.cpp
```diff
@@ -0,0 +1,140 @@
...
+                    HandleSpawnAIEntity(nullptr, m_teamID);
+                }
+                else
+                {
+                    SpawnAIEntity(m_teamID);
+                }
+            }
+        }
+    }
+#endif
```
AMZN-koppersr, line 103: If the `#if` and `#endif` are more than a few lines apart it's recommended to add the condition as a comment to the `#endif` like `#endif //defined(IMGUI_ENABLED)` as this makes it easier to read.

Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
```diff
@@ -97,15 +115,19 @@ namespace MultiplayerSample
...
+            aznumeric_cast<uint32_t>(CharacterAnimState::Jumping), wasdInput->m_jump);
+        GetNetworkAnimationComponentController()->ModifyActiveAnimStates().SetBit(
+            aznumeric_cast<uint32_t>(CharacterAnimState::Crouching), wasdInput->m_crouch);
 
         // Update orientation
         AZ::Vector3 aimAngles = GetSimplePlayerCameraComponentController()->GetAimAngles();
         aimAngles.SetZ(NormalizeHeading(aimAngles.GetZ() - wasdInput->m_viewYaw * cl_AimStickScaleZ));
         aimAngles.SetX(NormalizeHeading(aimAngles.GetX() - wasdInput->m_viewPitch * cl_AimStickScaleX));
-        aimAngles.SetX(NormalizeHeading(AZ::GetClamp(aimAngles.GetX(), -AZ::Constants::QuarterPi * 0.75f, AZ::Constants::QuarterPi * 0.75f)));
+        aimAngles.SetX(
```
AMZN-koppersr, line 129: Why is X set, retrieved and set again? Wouldn't it be better to store this in a local value and only set X once?

Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
```diff
@@ -97,15 +115,19 @@ namespace MultiplayerSample
...
+            aznumeric_cast<uint32_t>(CharacterAnimState::Sprinting), wasdInput->m_sprint);
+        GetNetworkAnimationComponentController()->ModifyActiveAnimStates().SetBit(
+            aznumeric_cast<uint32_t>(CharacterAnimState::Jumping), wasdInput->m_jump);
+        GetNetworkAnimationComponentController()->ModifyActiveAnimStates().SetBit(
+            aznumeric_cast<uint32_t>(CharacterAnimState::Crouching), wasdInput->m_crouch);
 
         // Update orientation
         AZ::Vector3 aimAngles = GetSimplePlayerCameraComponentController()->GetAimAngles();
         aimAngles.SetZ(NormalizeHeading(aimAngles.GetZ() - wasdInput->m_viewYaw * cl_AimStickScaleZ));
         aimAngles.SetX(NormalizeHeading(aimAngles.GetX() - wasdInput->m_viewPitch * cl_AimStickScaleX));
```
AMZN-koppersr, line 128: There's a bit of SIMD roundtripping going on here that can be avoided.

Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
jeremyong-az: Sorry this is just a formatting change for code that is unchanged from before

Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
jeremyong-az: (I'll defer this to the other folks on the networking team to look at :D)

Gem/Code/Source/AutoGen/NetworkAiComponent.AutoComponent.xml
```diff
@@ -0,0 +1,14 @@
...
+<Component
+    Name="NetworkAiComponent" 
+    Namespace="MultiplayerSample" 
+    OverrideComponent="true" 
+    OverrideController="false" 
+    OverrideInclude="Source/Components/NetworkAiComponent.h"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+    <NetworkProperty Type="bool" Name="Enabled" Init="false" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="false" IsPredictable="false" ExposeToEditor="true" ExposeToScript="false" GenerateEventBindings="false" Description="If enabled, this AI component overrides movement and camera components." />
+    <NetworkProperty Type="uint32_t" Name="OwningConnectionId" Init="0xffffffff" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" ExposeToScript="false" GenerateEventBindings="false" Description="Defines whether the host or a particular client should drive the AI of this component's parent entity. A max uint32_t value indicates that the server ticks the AI."/>
```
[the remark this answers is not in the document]
kberg-amzn: OwningConnectionId going over the wire has me on edge. I will read on to see how this is used!

Gem/Code/multiplayersample_files.cmake
```diff
@@ -15,9 +16,12 @@ set(FILES
     Source/AutoGen/NetworkRigidBodyComponent.AutoComponent.xml
     Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
     Source/AutoGen/SimplePlayerCameraComponent.AutoComponent.xml
+    Source/AutoGen/StressTestCOmponent.AutoComponent.xml
```
kberg-amzn, line 19: COmponent <= Component

Gem/Code/Source/AutoGen/NetworkAiComponent.AutoComponent.xml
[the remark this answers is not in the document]
kberg-amzn: Looks like it's not used by anything on the other side receiving replicated updates for the entity. In fact it doesn't really appear to be used at all. This is good, since the value would be meaningless to anyone but the original host (connectionId is specific to the INetworkInstance that owns it). If you really need this to be a network property, I might change it to Authority to Authority. Otherwise I might just make this a regular member variable on your derived component/controller or remove it entirely.

Gem/Code/Source/Components/SimplePlayerCameraComponent.cpp
```diff
@@ -6,6 +6,7 @@
  */
```
kberg-amzn, line 6: I might be tempted to make a separate AI prefab that simply doesn't contain this component.

jeremyong-az: Yea would be down for a change like that, it's actually how I started but the movement controller for the time being depended on it so I left it in to limit the scope of the change.

jeremyong-az:
Latest PR update adds this menu:

image

Gem/Code/Source/AutoGen/NetworkAiComponent.AutoComponent.xml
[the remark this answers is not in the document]
jeremyong-az: Will remove and circle back offline to understand how to execute this pattern

kberg-amzn (approved):
Looks good. I still have minor feedback but we can tackle remaining changes in a future PR.

Gem/Code/Source/Components/NetworkAiComponent.cpp
```diff
@@ -0,0 +1,134 @@
...
+    {
+        // TODO: Execute this tick only if this component is owned by this endpoint (currently ticks on server only)
+        m_remainingTimeMs -= deltaTime * 1000;
+
+        if (m_remainingTimeMs <= 0)
+        {
+            // Determine a new directive after 500 to 9500 ms
+            m_remainingTimeMs = m_lcg.GetRandomFloat() * (m_actionIntervalMaxMs - m_actionIntervalMinMs) + m_actionIntervalMinMs;
+            m_turnRate = 1.f / m_remainingTimeMs;
+            m_targetYawDelta = -movementController.m_viewYaw + (m_lcg.GetRandomFloat() * 2.f - 1.f);
```
lmbr-pip, line 51: nit: add description of intent as above

Gem/Code/Source/Components/NetworkAiComponent.cpp
```diff
@@ -0,0 +1,134 @@
...
+            m_targetPitchDelta = -movementController.m_viewPitch + (m_lcg.GetRandomFloat() - 0.5f);
+            m_action = static_cast<Action>(m_lcg.GetRandom() % static_cast<int>(Action::COUNT));
+            m_strafingRight = static_cast<bool>(m_lcg.GetRandom() % 2);
+        }
+
+        // Translate desired motion into inputs
+
+        // Interpolate the current view yaw and pitch values towards the desired values
+        movementController.m_viewYaw += 1000.f * m_turnRate * deltaTime * m_targetYawDelta;
```
lmbr-pip, line 60: nit: should 1000.f be a constexpr somewhere here?

Gem/Code/Source/Components/NetworkAiComponent.h
```diff
@@ -0,0 +1,70 @@
...
+#include <Source/AutoGen/NetworkAiComponent.AutoComponent.h>
+
+#include <AzCore/Math/Random.h>
+
+namespace MultiplayerSample
+{
+    class NetworkWeaponsComponentController;
+    class WasdPlayerMovementComponentController;
+
+    class NetworkAiComponent : public NetworkAiComponentBase
```
lmbr-pip, line 19: nit: prefer to have a description of class ie key responsibilities to aid maintainers/readers of code

Gem/Code/Source/Components/NetworkAiComponent.h
```diff
@@ -0,0 +1,70 @@
...
+        friend class StressTestComponentController;
+        void ConfigureAi(
+            float fireIntervalMinMs, float fireIntervalMaxMs, float actionIntervalMinMs, float actionIntervalMaxMs, uint64_t seed);
+
+        AZ::SimpleLcgRandom m_lcg;
+
+        // Our "AI" is really just a chaos monkey. Every N ms, we choose a cardinal direction to move towards,
+        // and flip coins to determine if we should shoot, or perform some other action.
+        float m_remainingTimeMs = 0.f;
+        float m_fireIntervalMinMs;
```
lmbr-pip, line 46: As these don't have default values, do we need to define a Ctor to set them up?

lmbr-pip (commented):
LG overall.

Gem/Code/Source/Components/NetworkAiComponent.cpp
jeremyong-az: This converts seconds to ms (since deltaTime in our engine is typically a float of seconds). Is there already a constant somewhere I should use, or would you prefer a locally defined constant?

Gem/Code/Source/Components/NetworkAiComponent.h
jeremyong-az: These are initialized by the `ConfigureAi` method above which is invoked in response to an RPC, but there's no harm in initializing them here as well

Gem/Code/Source/Components/NetworkAiComponent.h
```diff
@@ -0,0 +1,73 @@
...
+
+        void TickMovement(WasdPlayerMovementComponentController& movementController, float deltaTime);
+        void TickWeapons(NetworkWeaponsComponentController& weaponsController, float deltaTime);
+
+    private:
+        friend class StressTestComponentController;
+        void ConfigureAi(
+            float fireIntervalMinMs, float fireIntervalMaxMs, float actionIntervalMinMs, float actionIntervalMaxMs, uint64_t seed);
+
+        AZ::SimpleLcgRandom m_lcg;
```
AMZN-Olex, line 44: It would be nice to use future Network Random once we write ;)

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -458,4 +477,10 @@ namespace MultiplayerSample
     {
         ;
     }
-}
+
+    void NetworkWeaponsComponentController::UpdateAI()
+    {
+        float deltaTime = static_cast<float>(m_updateAI.TimeInQueueMs()) / 1000.f;
```
AMZN-Olex, line 483: Our standard is using aznumeric_cast<> instead of static_cast for numbers.

AMZN-Olex (approved):

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
jeremyong-az: There was a discussion in another channel actually where it turns out that aznumeric_cast doesn't do anything (the assert is commented I believe)

PR #63, opened 2021-09-10
Moving rewind out of WASD into NetworkCharacter::TryMove

AMZN-Gene:
Moving rewind out of WASD Character Controller into NetworkCharacter::TryMove. Anyone wanting to move would want to make sure they rewind other entities they potentially will come in contact with...

### DIFF
--- a/Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
+++ b/Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
@@ -115,16 +115,6 @@ namespace MultiplayerSample
         // Update velocity
         UpdateVelocity(*wasdInput);
 
-        // Ensure any entities that we might interact with are properly synchronized to their rewind state
-        if (IsAuthority())
-        {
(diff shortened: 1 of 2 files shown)


Gem/Code/Source/Components/WasdPlayerMovementComponent.cpp
```diff
@@ -115,16 +115,6 @@ namespace MultiplayerSample
         // Update velocity
         UpdateVelocity(*wasdInput);
 
-        // Ensure any entities that we might interact with are properly synchronized to their rewind state
-        if (IsAuthority())
-        {
-            const AZ::Aabb entityStartBounds = AZ::Interface<AzFramework::IEntityBoundsUnion>::Get()->GetEntityLocalBoundsUnion(GetEntity()->GetId());
```
SergeyAMZN, line 121: I guess we don't need #include <AzFramework/Visibility/EntityBoundsUnionBus.h> in this file anymore?

SergeyAMZN (approved):

PR #85, opened 2021-11-10
Add GameLift integration for the MultiplayerSample

junbo75:
Signed-off-by: Junbo Liang <junbo@amazon.com>

- Defined a separate server and client for the GameLift integration since the current project doesn't separate the client and server targets but the AWSGameLift gem does.

- Add a new cmake argument to build the GameLift specific server and client.

- Define basic client and server logic for tutorial/automation tests purpose. Add a few console commands for session management and matchmaking requests.

- Tested the server and client connection against GameLift (Have to remove all the unused server target like the atom stuff manually).

### DIFF
--- a/Gem/Code/CMakeLists.txt
+++ b/Gem/Code/CMakeLists.txt
@@ -5,6 +5,8 @@
 #
 #
 
+set(GAMELIFT_INTEGRATION OFF CACHE BOOL "Indicates if the game will be built with the GameLift integration features")
+
 ly_get_list_relative_pal_filename(pal_dir ${CMAKE_CURRENT_LIST_DIR}/Platform/${PAL_PLATFORM_NAME})
 
 ly_add_target(
@@ -60,11 +62,95 @@ ly_add_target(
             Gem::Atom_AtomBridge.Static
 )
 
+ly_add_target(
+    NAME MultiplayerSample.GameLiftServer.Static STATIC
+    NAMESPACE Gem
+    FILES_CMAKE
+        multiplayersample_gamelift_server_files.cmake
+    INCLUDE_DIRECTORIES
+        PRIVATE
+            Source
+            .
+        PUBLIC
+            Include
+    BUILD_DEPENDENCIES
+        PUBLIC
+            AZ::AzCore
+            AZ::AzFramework
+            AZ::AzNetworking
+            Gem::Multiplayer
+            Gem::AWSGameLift.Servers
+)
+
+ly_add_target(
+    NAME MultiplayerSample.GameLiftServer ${PAL_TRAIT_MONOLITHIC_DRIVEN_MODULE_TYPE}
+    NAMESPACE Gem
+    FILES_CMAKE
+        multiplayersample_gamelift_server_shared_files.cmake
+    INCLUDE_DIRECTORIES
+        PRIVATE
+            Source
+            .
+        PUBLIC
+            Include
+    BUILD_DEPENDENCIES
+        PRIVATE
+            Gem::MultiplayerSample.Static
+            Gem::MultiplayerSample.GameLiftServer.Static
+    RUNTIME_DEPENDENCIES
+            Gem::AWSGameLift.Servers
+)
+
+ly_add_target(
+    NAME MultiplayerSample.GameLiftClient.Static STATIC
+    NAMESPACE Gem
+    FILES_CMAKE
+        multiplayersample_gamelift_client_files.cmake
+    INCLUDE_DIRECTORIES
+        PRIVATE
+            Source
+            .
+        PUBLIC
+            Include
+    BUILD_DEPENDENCIES
+        PUBLIC
+            AZ::AzCore
+            AZ::AzFramework
+            Gem::AWSCore
+            Gem::AWSGameLift.Clients
+)
+
+ly_add_target(
+    NAME MultiplayerSample.GameLiftClient ${PAL_TRAIT_MONOLITHIC_DRIVEN_MODULE_TYPE}
+    NAMESPACE Gem
+    FILES_CMAKE
+        multiplayersample_gamelift_client_shared_files.cmake
+        ../../multiplayersample_asset_files.cmake
+    INCLUDE_DIRECTORIES
+        PRIVATE
+            Source
+            .
+        PUBLIC
+            Include
+    BUILD_DEPENDENCIES
+        PRIVATE
+            Gem::MultiplayerSample.Static
+            Gem::MultiplayerSample.GameLiftClient.Static
+    RUNTIME_DEPENDENCIES
+            Gem::AWSGameLift.Clients
+)
+
 # if enabled, MultiplayerSample is used by all kinds of applications
 ly_create_alias(NAME MultiplayerSample.Builders NAMESPACE Gem TARGETS Gem::MultiplayerSample)
 ly_create_alias(NAME MultiplayerSample.Tools    NAMESPACE Gem TARGETS Gem::MultiplayerSample)
-ly_create_alias(NAME MultiplayerSample.Clients  NAMESPACE Gem TARGETS Gem::MultiplayerSample)
-ly_create_alias(NAME MultiplayerSample.Servers  NAMESPACE Gem TARGETS Gem::MultiplayerSample)
+
+if(GAMELIFT_INTEGRATION)
+    ly_create_alias(NAME MultiplayerSample.Clients  NAMESPACE Gem TARGETS Gem::MultiplayerSample.GameLiftClient)
+    ly_create_alias(NAME MultiplayerSample.Servers  NAMESPACE Gem TARGETS Gem::MultiplayerSample.GameLiftServer)
+else()
+    ly_create_alias(NAME MultiplayerSample.Clients  NAMESPACE Gem TARGETS Gem::MultiplayerSample)
+    ly_create_alias(NAME MultiplayerSample.Servers  NAMESPACE Gem TARGETS Gem::MultiplayerSample)
+endif()
 
 ################################################################################
 # Gem dependencies
--- a/Gem/Code/Source/GameLift/GameLiftClientComponent.cpp
+++ b/Gem/Code/Source/GameLift/GameLiftClientComponent.cpp
@@ -1,0 +1,377 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project. For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzCore/Math/Uuid.h>
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/Serialization/EditContext.h>
+#include <AzCore/Serialization/EditContextConstants.inl>
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzFramework/Session/SessionConfig.h>
+
+#include "GameLiftClientComponent.h"
+#include <AWSGameLiftPlayer.h>
+#include <Request/AWSGameLiftAcceptMatchRequest.h>
+#include <Request/AWSGameLiftCreateSessionOnQueueRequest.h>
+#include <Request/AWSGameLiftCreateSessionRequest.h>
+#include <Request/AWSGameLiftJoinSessionRequest.h>
+#include <Request/AWSGameLiftSearchSessionsRequest.h>
+#include <Request/AWSGameLiftStartMatchmakingRequest.h>
+#include <Request/AWSGameLiftStopMatchmakingRequest.h>
+#include <ResourceMapping/AWSResourceMappingBus.h>
+
+namespace MultiplayerSample
+{
+    AZ_CVAR(bool, cl_joinaftersearch, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
+        "Whether to join the first available session after search automatically");
+    AZ_CVAR(bool, cl_acceptmatch, false, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
+        "Whether to accept the match automatically");
+
+    void GameLiftClientSystemComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<GameLiftClientSystemComponent, AZ::Component>()->Version(0);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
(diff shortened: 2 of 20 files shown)


Gem/Code/Source/GameLift/GameLiftClientComponent.cpp
```diff
@@ -0,0 +1,377 @@
...
+    void GameLiftClientSystemComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<GameLiftClientSystemComponent, AZ::Component>()->Version(0);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<GameLiftClientSystemComponent>(
+                      "GameLiftClientSystemComponent", "System Component for the GameLift client")
```
lmbr-pip, line 42: nit: Call out this component is the MPS GameLift client so you can easily tell GameLift gem vs project components ie MPSGameLiftClientSystemComponent.

Gem/Code/Source/GameLift/GameLiftClientComponent.h
```diff
@@ -0,0 +1,98 @@
...
+#include <AzFramework/Matchmaking/MatchmakingNotifications.h>
+
+#include <Request/AWSGameLiftRequestBus.h>
+#include <Request/AWSGameLiftSessionRequestBus.h>
+#include <Request/AWSGameLiftMatchmakingRequestBus.h>
+
+
+namespace MultiplayerSample
+{
+    class GameLiftClientSystemComponent
```
lmbr-pip, line 20: Consider calling this the MultiplayerGameLiftClientSystemComponent or MPSGameLiftClientSystemComponent to avoid any confusion with things in the GameLift gem.

Add comment about class description/responsibilities

Gem/Code/Source/GameLift/GameLiftClientComponent.h
```diff
@@ -0,0 +1,98 @@
...
+        static void GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent);
+
+    protected:
+        ////////////////////////////////////////////////////////////////////////
+        // AZ::Component interface implementation
+        void Init() override;
+        void Activate() override;
+        void Deactivate() override;
+
+        void HostSession(const AZ::ConsoleCommandContainer& consoleFunctionParameters);
```
lmbr-pip, line 43: nit: Consider adding brief comments to this method esp as you are providing functionality and registered console funcs

Gem/Code/Source/GameLift/GameLiftClientComponent.h
```diff
@@ -0,0 +1,98 @@
...
+        void OnMatchFailure() override;
+
+         ////////////////////////////////////////////////////////////////////////
+
+    private:
+        void AcceptMatch(bool accept = true);
+        void JoinSessionInternal(const AZStd::string& sessionId, const AZStd::string& playerId);
+
+        AZStd::string m_ticketId;
+        AZStd::string m_playerId;
```
lmbr-pip, line 96: nit: Add comment this is a locally generated player id, devs should feel free to replace this id with any other suitable id. It just uniquely identifies this player in a session to GameLift.

Gem/Code/Source/GameLift/GameLiftServerComponent.h
```diff
@@ -0,0 +1,67 @@
...
+#include <AzFramework/Session/SessionNotifications.h>
+
+namespace AzNetworking
+{
+    class INetworkInterface;
+}
+
+namespace MultiplayerSample
+{
+    class GameLiftServerSystemComponent
```
lmbr-pip, line 22: nit: Same as above, consider MultiplayerGameLiftSampleServer... or MPSGameLiftServer...

Class needs comment wrt to description/responsibilities.

Gem/Code/Source/GameLift/MultiplayerSampleGameLiftServerModule.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "GameLift/GameLiftServerComponent.h"
+
+namespace MultiplayerSample
+{
+    class GameLiftServerModule
+        : public AZ::Module
+    {
+    public:
+        AZ_RTTI(GameLiftServerModule, "{C70DF768-3BB1-46BC-9F96-DE86B53610D2}", AZ::Module);
+        AZ_CLASS_ALLOCATOR(GameLiftServerModule, AZ::SystemAllocator, 0);
```
lmbr-pip, line 22: nit: See comments around MPS vs GameLift naming

lmbr-pip (commented):
LG, overall.

"Test the server and client connection against GameLift (Have to remove all the unused server target like the atom stuff manually)." < -- assume you meant tested, otherwise this reads like you still need to test.

junbo75:
> LG, overall.
> 
> "Test the server and client connection against GameLift (Have to remove all the unused server target like the atom stuff manually)." < -- assume you meant tested, otherwise this reads like you still need to test.

Yes, I mean "tested" here :). Just updated the description.

Gem/Code/Source/GameLift/GameLiftClientComponent.cpp
```diff
@@ -0,0 +1,377 @@
...
+        if (!sessionIds.empty())
+        {
+            sessionIds = sessionIds.substr(0, sessionIds.size() - 2);
+        }
+
+        AZ_TracePrintf("GameLiftClientSystemComponent", "SearchSessionsAsync complete with result: %s",
+            AZStd::string::format("[%s]", sessionIds.c_str()).c_str());
+
+        const auto console = AZ::Interface<AZ::IConsole>::Get();
+        bool joinaftersearch = false;
```
AMZN-puvvadar, line 269: NP: joinAfterSearch to meet naming standards

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+        {
+            m_waitEvent.try_acquire_for(AZStd::chrono::seconds(30));
+
+            if (!m_backfillComplete || !HasEmptySlots())
+            {
+                // The previous backfill request hasn't been completed or there's no available slots for backfill
+                continue;
+            }
+
+            AZ_TracePrintf("MultiplayerSample", "Initiate multiplayer sample server manual backfill");
```
AMZN-puvvadar, line 189: In general, consider switching your AZ_TracePrintf usages to categorized AZLOGs or similar

Gem/Code/Source/GameLift/GameLiftClientComponent.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        m_ticketId = "";
+        AWSGameLift::AWSGameLiftMatchmakingEventRequestBus::Broadcast(
+            &AWSGameLift::AWSGameLiftMatchmakingEventRequestBus::Events::StopPolling);
+    }
+
+    void GameLiftClientSystemComponent::OnMatchAcceptance()
+    {
+        const auto console = AZ::Interface<AZ::IConsole>::Get();
+        bool acceptmatch = false;
```
AMZN-puvvadar, line 333: NP: acceptMatch

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+            AZStd::string matchmakingTicketId = AZ::Uuid::CreateRandom().ToString<AZStd::string>(false, false);
+            AZStd::vector<AWSGameLift::AWSGameLiftPlayer> players;
+            m_backfillComplete = false;
+
+            bool result = false;
+            AWSGameLift::AWSGameLiftServerRequestBus::BroadcastResult(result,
+                &AWSGameLift::AWSGameLiftServerRequestBus::Events::StartMatchBackfill, matchmakingTicketId, players);
+            if (!result)
+            {
+                m_backfillComplete = true;
```
AMZN-puvvadar, line 200: Does the semaphore need a release here?

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    void GameLiftServerSystemComponent::OnUpdateSessionBegin(
+        const AzFramework::SessionConfig& sessionConfig, const AZStd::string& updateReason)
+    {
+        AZ_UNUSED(sessionConfig);
+        if (updateReason == "BACKFILL_FAILED" || updateReason == "BACKFILL_TIMED_OUT" || updateReason == "BACKFILL_CANCELLED")
+        {
+            m_backfillComplete = true;
+            m_waitEvent.release();
```
AMZN-puvvadar, line 123: Where is this released on success? I'm not sure what the callback is for completed backfill

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    void GameLiftServerSystemComponent::BackfillProcess()
+    {
+        while (!m_backfillProcessTerminated)
+        {
+            m_waitEvent.try_acquire_for(AZStd::chrono::seconds(30));
+
+            if (!m_backfillComplete || !HasEmptySlots())
```
AMZN-puvvadar, line 183: So the idea is to poll and check for empty slots on a side thread? There are events on the IMultiplayer interface you can subscribe to that let you react to slot availability vs polling. If those aren't quite sufficient, IMultiplayer should be updated such that they are

AMZN-puvvadar:
This is fairly significant set of changes. It'll take some time for me to fully digest. I've left comments on some nitpicks and concerns around threading in the server system component.

An immediate question is we have MultiplayerSystemComponent implementing ISessionHandlingClientRequests and SessionNotificationBus. What about the client/server split in AWSGameLift required implementing separate System Components in Multiplayer Sample vs making use of that machinery in MultiplayerSystemComponent?

Gem/Code/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@
 #
 #
 
+set(GAMELIFT_INTEGRATION OFF CACHE BOOL "Indicates if the game will be built with the GameLift integration features")
```
AMZN-Olex, line 8: I don't like this switch in the body of the project.
I would like to keep it cloud agnostic.
Please move it to a GameLift gem instead and export it to here. See sig-build for help with this.

AMZN-Olex (changes requested):

AMZN-Olex:
Is there any way to make this cloud option be runtime instead of compile time? This compile switching will make it very hard to test it in automatic fashion since you cannot reuse one build to test different cloud systems.

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+        AZ_TracePrintf("MultiplayerSample", "Multiplayer sample server manual backfill has completed");
+
+        m_backfillComplete = true;
+    }
+
+    void GameLiftServerSystemComponent::StartBackfillProcess()
+    {
+        bool manualBackfill = false;
+        const auto console = AZ::Interface<AZ::IConsole>::Get();
+        if (console->GetCvarValue("sv_manualBackfill", manualBackfill) != AZ::GetValueResult::Success)
```
kberg-amzn, line 138: You don't have to do any of this if the cvar is declared within the same translation unit. Just use sv_manualBackfill as you would any normal static variable. That AZ:CVAR macro will expand to something similar to `static bool sv_manualBackfill = false;`, just with some extra logic for console binding and thread safety.

That means this whole block can just be

```
void GameLiftServerSystemComponent::StartBackfillProcess()
    {
        if (!sv_manualBackfill|| !m_backfillProcessTerminated)
        {
            return;
        }
```

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
```diff
@@ -0,0 +1,211 @@
...
+        AZ_TracePrintf("MultiplayerSample", "Start multiplayer sample server manual backfill");
+        m_backfillProcessTerminated = false;
+        m_backfillThread = AZStd::thread(AZStd::bind(&GameLiftServerSystemComponent::BackfillProcess, this));
+    }
+
+    void GameLiftServerSystemComponent::StopBackfillProcess()
+    {
+        bool manualBackfill = false;
+        const auto console = AZ::Interface<AZ::IConsole>::Get();
+        if (console->GetCvarValue("sv_manualBackfill", manualBackfill) != AZ::GetValueResult::Success)
```
kberg-amzn, line 158: Same deal, this collapses down to using sv_manualBackfill directly.

Gem/Code/CMakeLists.txt
```diff
@@ -60,11 +62,95 @@ ly_add_target(
             Gem::Atom_AtomBridge.Static
 )
 
+ly_add_target(
```
kberg-amzn, line 65: It seems a bit weird that these modules aren't defined in the gamelift gem itself with a config file or settings registry to drive most game specific behaviour? Then the integration into mpsample would be vastly lighter, the gamelift gem would provide significantly more value out of the box, and also for customers wanting to bring the gamelift gem into their own project they wouldn't be copy-pasting so much code out of mpsample.

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
junbo75: Yeah, that's the idea here. GameLift suggests to initiate a backfill request whenever a matched game has one or more empty player slots. I'll check the IMultiplayer interface on it.

Gem/Code/Source/GameLift/GameLiftServerComponent.cpp
junbo75: The idea here is to unblock the acquirer when backfill fails so that the side thread will retry the backfill immediately. I think we can just wait for the 30s time period to pass (m_waitEvent.try_acquire_for(AZStd::chrono::seconds(30))) on success and start another backfill request if there're empty slots.

If it's possible to subscribe to the IMultiplayer interface and react to slot availability, then we may not need the semaphore at all. I'll look into the interface.

AMZN-puvvadar: To be clear, if the IMultiplayer interface isn't meeting that need you should feel free to update it so that it does. Locking mechanisms are generally costly so eliminating that plus the wait/spin would be positive.

junbo75:
Close the PR and will revisit it once we have clear client and server split in MPSS.

Issue #97, opened 2021-12-01
[bug] MultiplayerSample not building on Linux due to compiler detection issues


lmbr-pip:
Using the stablization branch installer project manager, seeing

```
CMake Error at /opt/O3DE/21.11/cmake/Platform/Linux/PAL_linux.cmake:34 (message):
  Compiler GNU not supported in Linux
Call Stack (most recent call first):
  /opt/O3DE/21.11/cmake/PAL.cmake:248 (include)
  /opt/O3DE/21.11/CMakeLists.txt:29 (include)
-- Configuring incomplete, errors occurred!
See also "/home/ANT.AMAZON.COM/rrobc/O3DE/Projects/o3de-multiplayersample/build/linux/CMakeFiles/CMakeOutput.log".
```
[CMakeOutput.log](https://github.com/o3de/o3de-multiplayersample/files/7636817/CMakeOutput.log)

Need to update cmake files to support latest compiler detection changes: https://github.com/o3de/o3de/pull/5376

PR #98, opened 2021-12-01
Adapts to new template

amznestebanpapp:
Closes Issue #97

Signed-off-by: Esteban Papp <81431996+amznestebanpapp@users.noreply.github.com>

### DIFF
--- a/cmake/Platform/Linux/CompilerSettings_linux.cmake
+++ b/cmake/Platform/Linux/CompilerSettings_linux.cmake
@@ -1,0 +1,34 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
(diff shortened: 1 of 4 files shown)


lmbr-pip (approved):
Does this other plaforms mac, android etc to have a file like  cmake/Platform/Linux/CompilerSettings_linux.cmake ?

cmake/Platform/Linux/CompilerSettings_linux.cmake
```diff
@@ -0,0 +1,34 @@
+#
```
lmbr-pip, line 1: nit: Should this have # {BEGIN_LICENSE} etc

amznestebanpapp: it should not, and the one above should not have it neither, is just from the template, will remove the one from above

AMZN-Olex (approved):

PR #102, opened 2021-12-15
Add server build command to README

allisaurus:
Notes cmds to build ServerLauncher target prior to running it. Also fixes a typo in link to `cvar` tutorial.

Signed-off-by: allisaurus <34254888+allisaurus@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,17 @@ If these cfg files are not present, create them as they will be used to when lau
 
 #### Running the Server
 
-A server can be run as follows
+First, make sure the server target has been built
+
+```shell
+# example command for Project-centric approach
+> cmake --build c:/path/to/o3de-multiplayersample/build --target Editor MultiplayerSample.ServerLauncher --config profile -- /m /nologo
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,15 +101,25 @@ If these cfg files are not present, create them as they will be used to when lau
 
 #### Running the Server
 
-A server can be run as follows
+First, make sure the server target has been built
```
lmbr-pip, line 104: As this is in the running the server section, this information should probably migrate to ### Step 3. Configure and build

Otherwise we are starting to cross pollinate a little.

allisaurus: tbh I waivered between this and putting it there; I guessed maybe its omission meant that it's more situational when you'd need to build & run the server (vs. the client/GameLauncher)? But if most folks are going to want to run & build both, then agreed Step 3 makes more sense.

lmbr-pip (approved):

AMZN-Olex (approved):

PR #106, opened 2022-01-21
Add IMultiplayerSpawner impl for MPS

AMZN-puvvadar:
A round robin implementation of spawner logic based on requirements outlined in https://github.com/o3de/o3de/pull/7060.

In order to register with spawning, an entity needs to add a NetworkPlayerSpawnerComponent and assign a networked Spawnable to the component. The RoundRobinSpawner will rotate through all available spawners as requests come in and return the assigned Asset and the transform of the spawner entity.

This change adds four spawn points to SampleBase pointed at the Player spawnable.


kberg-amzn:
This is good, but what are your thoughts on making an actual player spawner component that we can place on an entity with a prefab asset editor property that a user can bind to anything they want. It's still an incredibly trivial implementation of your interface, but it would kill the while player.prefab being hardcoded in a cvar and give designers explicit control over what and where players get spawned.. really good setup for the upcoming sample initiative so probably well worth the effort.

AMZN-puvvadar:
> This is good, but what are your thoughts on making an actual player spawner component that we can place on an entity with a prefab asset editor property that a user can bind to anything they want. It's still an incredibly trivial implementation of your interface, but it would kill the while player.prefab being hardcoded in a cvar and give designers explicit control over what and where players get spawned.. really good setup for the upcoming sample initiative so probably well worth the effort.

It's something I want to get to though I'm focusing on just getting the interface defined out. Like you're alluding to, the idea is the IMultiplayerSpawner implementation could route the request off to a system based around a component type we could attach to entities.

AMZN-Gene (approved):

Gem/Code/Source/Components/NetworkPlayerSpawnerComponent.h
```diff
@@ -1,8 +1,31 @@
...
+#include <Source/AutoGen/NetworkPlayerSpawnerComponent.AutoComponent.h>
+
+namespace MultiplayerSample
+{
+
+    class NetworkPlayerSpawnerComponent : public NetworkPlayerSpawnerComponentBase
+    {
+    public:
+        AZ_MULTIPLAYER_COMPONENT(
+            MultiplayerSample::NetworkPlayerSpawnerComponent,
```
kberg-amzn, line 21: Don't think you need to duplicate the namespaces in this macro..?

kberg-amzn (commented):
So if I understand the intended architecture, here are some minor suggestions that might make this a little more flexible.
- Make your RoundRobinSpawner which conforms to the IPlayerSpawner interface.
- Have the NetworkPlayerSpawnerComponent register itself to the IPlayerSpawner interface as you currently do.
- Move the cvar to the NetworkPlayerSpawnerComponent for now, since the spawner instance should probably be responsible for allocating and deallocating the actual prefab.
- The NetworkPlayerSpawnerComponent can use its own entity transform as the location to spawn a player prefab at.
- Move the selection of a spawner to the RoundRobin implementation, which does exactly that, iterates through the set of registered spawners every time player connect is invoked.

This has the nice property that all the round robin logic is fully contained inside the round robin class and fully decoupled from the details of the spawner component(s) that are registered. The MultiplayerSampleSystemComponent only has two one-line hooks into the IPlayerSpawner interface.

Future work, but since the NetworkPlayerSpawnerComponent is only responsible for vending and destroying player prefabs now, we can really easily replace that single cvar with a component specific property that points at a prefab assetId to spawn.

Now when Starr or Jonny want to make a multiplayer sample level, they pop open the editor, create a few entities where they want players to spawn and add the network spawner components to them, then for each spawner component they click the 'choose asset' widget which will pop open a list of prefabs to select from. This allows them to control exactly which prefabs are used to spawn players, as well as exactly where in the map those prefabs will get spawned. It's all configurable from inside the editor with no engineering intervention.

Otherwise you're doing some really great work here! :thumbsup:

AMZN-Olex:
> This is good, but what are your thoughts on making an actual player spawner component that we can place on an entity with a prefab asset editor property that a user can bind to anything they want. It's still an incredibly trivial implementation of your interface, but it would kill the while player.prefab being hardcoded in a cvar and give designers explicit control over what and where players get spawned.. really good setup for the upcoming sample initiative so probably well worth the effort.

@kberg-amzn Rajiv and I spoke off-line. We can use the spawner from my PR here: [Gem/Code/Source/Components/PerfTest/NetworkPrefabSpawnerComponent.h](https://github.com/o3de/o3de-multiplayersample/pull/110/files#diff-7a5297d3a465cb0071278d3cc5476051d78e9eebe491d0c69749773c2e70c58d)

AMZN-puvvadar:
Ultimately I ended up using a design similar to @AMZN-Olex but separate given it's been specialized for the PerfTest level. My component takes an Asset and routes through NetworkEntityManager to spawn a prefab. It seems like there's some outstanding work on the NetworkEntityManager side that could see this streamlined a _bit_ more but it's in good shape.

Gem/Code/Source/Spawners/RoundRobinSpawner.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Multiplayer
+{
+    struct EntityReplicationData;
+    using ReplicationSet = AZStd::map<ConstNetworkEntityHandle, EntityReplicationData>;
+} // namespace Multiplayer
+
+namespace MultiplayerSample
+{
+    class RoundRobinSpawner
```
AMZN-Olex, line 26: When is this created? I'm curious how the dependency order works between RoundRobinSpawner and NetworkPlayerSpawnerComponent.

AMZN-puvvadar: Activation of MPSampleSystemComponent. It should precede NetworkPlayerSpawnerComponent Activation reliably.

AMZN-puvvadar:
Latest change fixes an issue where NetworkSimplePlayerCameraComponent would blow out initial rotation by assuming zeroes for initial aim angles vs syncing with transform rotation.

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -375,7 +375,7 @@ namespace MultiplayerSample
                 }
 
                 // Validate the proposed start position is reasonably close to the related bone
-                float startPositionClampRange = .5f;
+                float startPositionClampRange = 1.f;
```
AMZN-Olex, line 378: This really is starting to look like this should be a cvar.

AMZN-Olex (approved):
I'm good with this. I just want a cvar in the comment I mentioned.

Gem/Code/Source/Components/NetworkSimplePlayerCameraComponent.cpp
```diff
@@ -72,13 +72,20 @@ namespace MultiplayerSample
         {
             const AZ::Quaternion targetRotation = AZ::Quaternion::CreateRotationZ(GetCameraYaw()) * AZ::Quaternion::CreateRotationX(GetCameraPitch());
             const AZ::Quaternion currentRotation = m_activeCameraEntity->GetTransform()->GetWorldTM().GetRotation();
-            const AZ::Quaternion aimRotation = m_syncOrientationImmediate ?
-                targetRotation : currentRotation.Slerp(targetRotation, cl_cameraBlendSpeed).GetNormalized();
+            AZ::Quaternion aimRotation;
+            if(m_syncOrientationImmediate)
```
AMZN-Olex, line 76: Are you seeing weird camera movements upon spawning otherwise?

AMZN-puvvadar: Yes, you'll see the camera slerp from origin to the intended initial orientation. It's honestly not a bad effect but shouldn't be default behavior.

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
AMZN-puvvadar: I suppose it might as well. This change was because I did not account for the range change while jumping but if the model changes it'll be worth adjusting.

Gem/Code/Source/AutoGen/NetworkPlayerSpawnerComponent.AutoComponent.xml
```diff
@@ -3,12 +3,16 @@
 <Component
     Name="NetworkPlayerSpawnerComponent" 
     Namespace="MultiplayerSample" 
-    OverrideComponent="false" 
+    OverrideComponent="true" 
     OverrideController="false" 
     OverrideInclude="Source/Components/NetworkPlayerSpawnerComponent.h"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
+	<Include File="AzCore/Asset/AssetSerializer.h"/>
```
kberg-amzn, line 11: tabs

Gem/Code/Source/Components/NetworkSimplePlayerCameraComponent.cpp
```diff
@@ -24,6 +24,11 @@ namespace MultiplayerSample
 
     void NetworkSimplePlayerCameraComponentController::OnActivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
     {
+	    // Synchronize aim angles with initial transform
```
kberg-amzn, line 27: Tab

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
kberg-amzn: +1 on the cvar.. any time you have a static magic number like this, may as well make it tunable.

kberg-amzn (approved):
Nice and clean. 👍

PR #107, opened 2022-01-22
Cmake Update: Remove Depreciated Use of ly_get_list_relative_pal_filename

AMZN-Gene:
Remove depreciated use of ly_get_list_relative_pal_filename; replaced with o3de_pal_dir

Tested by reconfiguring and building MultiplayerSample on Windows and seeing the warnings go away.

### DIFF
--- a/Gem/Code/CMakeLists.txt
+++ b/Gem/Code/CMakeLists.txt
@@ -5,7 +5,7 @@
 #
 #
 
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm (approved):

lmbr-pip:
Can we call out platforms change was tested on? Assume Windows only.

AMZN-Gene:
> Can we call out platforms change was tested on? Assume Windows only.

comment fixed!

PR #111, opened 2022-02-09
Fix various issues leading to desyncs while shooting

AMZN-puvvadar:
This change addresses a few issues causing desyncs in MultiplayerSample when firing the player's weapon

1. Replicates the start position of the shot as animation bones are not synced making the fire position unreliable
2. When filtering rewindable objects, prevent checking static rigid bodies as they're expected to be stationary so their rewindable frame shouldn't be relevant. This fixes the floor being filtered out of collision logic.

https://github.com/o3de/o3de/pull/7534/files addressed changes in the core around NetworkTime and Rewindables.

### DIFF
--- a/Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
@@ -16,7 +16,7 @@
 
     <NetworkInput Type="bool" Name="Draw" Init="false" />
     <NetworkInput Type="WeaponActivationBitset" Name="Firing"  Init="" />
-	<NetworkInput Type="AZ::Vector3" Name="FireTranslation" Init="" Container="Array" Count="MaxWeaponsPerComponent"/>
+	<NetworkInput Type="AZ::Vector3" Name="ShotStartPosition" Init="" />
 
     <NetworkProperty Type="FireParams"  Name="ActivationParams" Init=""  Container="Array" Count="MaxWeaponsPerComponent" ReplicateFrom="Authority" ReplicateTo="Client"     IsPublic="false" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" GenerateEventBindings="false" Description="Parameters for the current weapon activation" />
     <NetworkProperty Type="uint8_t"     Name="ActivationCounts" Init="0" Container="Array" Count="MaxWeaponsPerComponent" ReplicateFrom="Authority" ReplicateTo="Client"     IsPublic="false" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" GenerateEventBindings="true"  Description="The number of activations" />
--- a/Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
@@ -16,6 +16,7 @@
 
     <NetworkInput Type="bool" Name="Draw" Init="false" />
     <NetworkInput Type="WeaponActivationBitset" Name="Firing"  Init="" />
+	<NetworkInput Type="AZ::Vector3" Name="FireTranslation" Init="" Container="Array" Count="MaxWeaponsPerComponent"/>
 
     <NetworkProperty Type="FireParams"  Name="ActivationParams" Init=""  Container="Array" Count="MaxWeaponsPerComponent" ReplicateFrom="Authority" ReplicateTo="Client"     IsPublic="false" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" GenerateEventBindings="false" Description="Parameters for the current weapon activation" />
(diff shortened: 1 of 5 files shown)


Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
```diff
@@ -16,6 +16,7 @@
 
     <NetworkInput Type="bool" Name="Draw" Init="false" />
     <NetworkInput Type="WeaponActivationBitset" Name="Firing"  Init="" />
+	<NetworkInput Type="AZ::Vector3" Name="FireTranslation" Init="" Container="Array" Count="MaxWeaponsPerComponent"/>
```
kberg-amzn, line 19: Indentation

kberg-amzn (approved):
Looking good 👍

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -91,24 +91,10 @@ namespace MultiplayerSample
 
     void NetworkWeaponsComponent::ActivateWeaponWithParams(WeaponIndex weaponIndex, WeaponState& weaponState, const FireParams& fireParams, bool validateActivations)
     {
-        const uint32_t weaponIndexInt = aznumeric_cast<uint32_t>(weaponIndex);
```
kberg-amzn, line 94: Not necessarily something to do in this change, but we might want a cvar that caps the max positional and orientational delta from current bone state between the input and the animation system. If we exceed the tolerance we can just throw away the input value and use the animation state.

AMZN-puvvadar: Yeah I intend to do just that before I pull this out of draft. A lot of the Weapon firing logic in general needs a good look.

Gem/Code/Source/AutoGen/NetworkWeaponsComponent.AutoComponent.xml
lmbr-pip: nit: Can we think of a better name than FireTranslation, this is a translation in reference to what? I don't have a good suggestion, but this is definitely a name I would have to look up in documentation based on current naming.

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -335,6 +321,21 @@ namespace MultiplayerSample
 
         weaponInput->m_draw = m_weaponDrawn;
         weaponInput->m_firing = m_weaponFiring;
+
+        uint32_t weaponIndexInt = 0;
+        if (weaponInput->m_firing.GetBit(weaponIndexInt))
+        {
+            // Temp hack for weapon firing due to late ebus binding in 1.14
```
lmbr-pip, line 328: Unsure what comment means, what is 1.14 in this context? LY 1.14? In which case is this a hack? Is it better to add something like "Add compensation for weapon firing.."

AMZN-puvvadar: Honestly I'm unsure either, this block of code was moved and I lack the context. I'll see if I can find out.

AMZN-puvvadar: As far as I can tell, this comment isn't relevant anymore so I've removed it.

lmbr-pip (approved):

AMZN-Gene (approved):

Gem/Code/Source/Components/NetworkWeaponsComponent.cpp
```diff
@@ -362,7 +363,25 @@ namespace MultiplayerSample
...
+                int32_t boneIdx = GetNetworkAnimationComponentController()->GetParent().GetBoneIdByName(fireBoneName);
+
+                AZ::Transform fireBoneTransform;
+                if (!GetNetworkAnimationComponentController()->GetParent().GetJointTransformById(boneIdx, fireBoneTransform))
+                {
+                    AZLOG_WARN("Failed to get transform for fire bone joint Id %u", boneIdx);
+                }
+
+                // Validate the proposed start position is reasonably close to the related bone
+                float startPositionClampRange = .5f;
```
AMZN-Olex, line 378: Should this be configurable? Or is this within normalized space already?

Gem/Code/Source/Weapons/SceneQuery.cpp
```diff
@@ -79,8 +80,9 @@ namespace MultiplayerSample
             auto ignoreEntitiesFilterCallback =
                 [&filter, networkEntityManager](const AzPhysics::SimulatedBody* body, [[maybe_unused]] const Physics::Shape* shape)
             {
-                // Exclude the bodies from another rewind frame
-                if ((filter.m_rewindFrameId != Multiplayer::InvalidHostFrameId)
+                // Exclude the dynamic bodies from another rewind frame
+                if (filter.m_rewindFrameId != Multiplayer::InvalidHostFrameId
+                    && body->GetNativeType() != PhysX::NativeTypeIdentifiers::RigidBodyStatic
```
SergeyAMZN, line 85: It may be better to use `&&  (body->GetFrameId() != AzPhysics::SimulatedBody::UndefinedFrameId)` instead of the NativeType check in case of there are non-static rigid bodies that are intentionally not rewound

SergeyAMZN (approved):

PR #115, opened 2022-02-23
Added a workaround switch to avoid issues with deactivating network entities.

AMZN-Olex:
"Respawn Enabled", if true will delete old entities, and spawn new ones. Otherwise, will stop spawning once the desired maximum count of instances is reached.

Signed-off-by: AMZN-Olex <5432499+AMZN-Olex@users.noreply.github.com>

### DIFF
--- a/Gem/Code/Source/AutoGen/NetworkTestSpawnerComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkTestSpawnerComponent.AutoComponent.xml
@@ -12,6 +12,7 @@
 	<ComponentRelation Constraint="Required" HasController="false" Name="NetworkPrefabSpawnerComponent" Namespace="MultiplayerSample" Include="Source/Components/PerfTest/NetworkPrefabSpawnerComponent.h" />
 
 	<ArchetypeProperty Type="bool" Name="Enabled" Init="true" ExposeToEditor="true" Description="Enables spawning of test prefabs." />
+	<ArchetypeProperty Type="bool" Name="RespawnEnabled" Init="false" ExposeToEditor="true" Description="Deletes old instances and spawns new ones when at the maximum live count." />
 	<ArchetypeProperty Type="int" Name="MaxLiveCount" Init="100" ExposeToEditor="true" Description="Maximum objects to keep alive, will delete older objects when the count goes above this value." />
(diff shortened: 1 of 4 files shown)


Gem/Code/Source/AutoGen/NetworkTestSpawnerComponent.AutoComponent.xml
```diff
@@ -12,6 +12,7 @@
 	<ComponentRelation Constraint="Required" HasController="false" Name="NetworkPrefabSpawnerComponent" Namespace="MultiplayerSample" Include="Source/Components/PerfTest/NetworkPrefabSpawnerComponent.h" />
 
 	<ArchetypeProperty Type="bool" Name="Enabled" Init="true" ExposeToEditor="true" Description="Enables spawning of test prefabs." />
+	<ArchetypeProperty Type="bool" Name="RespawnEnabled" Init="false" ExposeToEditor="true" Description="Deletes old instances and spawns new ones when at the maximum live count." />
```
lmbr-pip, line 15: As maximum live count is a cvar, could we reference it by name? "when MaxiumumLiveCount is reached." Also is there anything folks should know about how instances are replaced? Are all replaced at once? Do we start replacing the oldest first? How does this function?

Gem/Code/Source/Components/PerfTest/NetworkTestSpawnerComponent.cpp
```diff
@@ -26,7 +26,10 @@ namespace MultiplayerSample
 
     void NetworkTestSpawnerComponentController::OnActivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
     {
-        m_tickEvent.Enqueue(AZ::TimeMs{ 0 }, true);
+        if (GetParent().GetNetworkPrefabSpawnerComponent())
```
lmbr-pip, line 29: nit: can we add a comment to why now on enqueue tick if prefab spawner?

lmbr-pip (commented):
LG.

Can you include how tested in your PR description? What platforms you checked and steps taken to ensure this change was good.

AMZN-Gene (approved):

kberg-amzn (approved):

PR #116, opened 2022-02-24
Add editor.cfg to enable local server from editor

allisaurus:
While running through [this tutorial](https://www.o3de.org/docs/learning-guide/tutorials/multiplayer/first-multiplayer-component/) with @AMZN-Gene, discovered local server launch from within the editor wasn't enabled by default. Feels like it should be for this project, so this change adds an `editor.cfg` file with `editorsv_enabled=true` and updates the README accordingly. 

NOTE: the [CVAR](https://www.o3de.org/docs/user-guide/programming/cvars/) link doesn't actually contain any content, so I removed it. (and created [this issue](https://github.com/o3de/o3de.org/issues/1441))

## Testing

1. Run the editor
2. Type `editorsv_enabled` in the console
3. the output should reflect this value is set to "true"
image

### DIFF
--- a/editor.cfg
+++ b/editor.cfg
@@ -1,0 +1,1 @@
+editorsv_enabled=true
--- a/README.md
+++ b/README.md
@@ -108,7 +108,9 @@ MultiplayerSample.ServerLauncher.exe --console-command-file=server.cfg
 
 #### Running the Server in the Editor
 
-Refer to the O3DE document [Test Multiplayer Games in the O3DE Editor](https://o3de.org/docs/user-guide/gems/reference/multiplayer/multiplayer-gem/test-in-editor/), to set up required console variables (cvar) to support play in editor with servers. Ensure you configure ```editorsv_enabled``` and ```editorsv_launch``` as required. See the [Console Variable Tutorial](https://o3de.org/docs/user-guide/engine/cvars) for more details on setting and using cvars.
+By default, launching a local server from the editor during Play Mode is enabled. To disable this behavior, update the `editorsv_enabled` value in the `editor.cfg` file to `false`.
+
+Refer to the O3DE document [Test Multiplayer Games in the O3DE Editor](https://o3de.org/docs/user-guide/gems/reference/multiplayer/multiplayer-gem/test-in-editor/) for the complete list of console variables (cvar) which support play in the editor with servers.
 
 
 #### Running the Client


AMZN-puvvadar (approved):

editor.cfg
```diff
@@ -0,0 +1 @@
+editorsv_enabled=true
```
AMZN-Gene, line 1: Is there any benefit to adding this to the Registry/commands.editor.setreg instead of cfg?

{
    "Amazon":
    {
        "AzCore":
        {
            "Runtime":
            {
                "ConsoleCommands":
                {
                    "editorsv_enabled": "true",
                }
            }
        }
    }
}

AMZN-Gene (approved):

editor.cfg
allisaurus: discussed offline: due to precedence for using other .cfg files in this repo, going to keep it like this for now.

kberg-amzn: What even loads editor.cfg though? I think that's a legacy load-path.. the only actual .cfg that we load at this point is I think user.cfg, and the server and client cfg files have to be explicitly loaded via command-line parameters.

allisaurus: So while testing I first added this CVAR to a `user.cfg`, but decided to go with a more specifically named file to further disambiguate it's contents from the `game.cfg` (adding `editorsv_enabled` there also seems to work, but I didn't want to entangle it with whatever purpose that file was created)

If it'd be better to put this (and perhaps the contents of the `game.cfg`?) into _Registry/commands.editor.setreg_ I'm happy to do so, it just seemed like `.cfg` files were the established way to go given we had 3 of them already.

allisaurus: @kberg-amzn After testing with a goofier .cfg file name, confirmed we're not arbitrarily loading any & all .cfg files from the project root (which was kind of my assumption 😬). 

It looks like [`editor.cfg`](https://github.com/o3de/o3de/blob/development/Code/Editor/CryEdit.cpp#L1283) (and [`game.cfg`](https://github.com/o3de/o3de/blob/development/Code/Legacy/CrySystem/SystemInit.cpp#L341)!) are being loaded by some old _Cry_ code, whose original purpose I'm reasonably sure didn't have to do with arbitrary local server cvars soooooo I am going to submit a follow up PR to move this into a .setreg file, as suggested by @AMZN-Gene 

Thanks for the tip that .cfg file values shouldn't just be loading from thin air!

PR #126, opened 2022-03-15
Editor server to set to hidden and RHI to null mode for faster startup

AMZN-Olex:
Signed-off-by: AMZN-Olex <5432499+AMZN-Olex@users.noreply.github.com>

### DIFF
--- a/editor.cfg
+++ b/editor.cfg
@@ -1,2 +1,4 @@
 editorsv_enabled=true
 ed_keepEditorAlive=true
+editorsv_hidden=true
+editorsv_rhi_override=null


AMZN-Gene (approved):
I’ll approve technically, but id question if new users wouldn’t want to see the server start up. I found it really useful, not only for seeing “oh something else started”, but to see the entities moving in both Server and client.
There really is an immediate feeling of “something different” when playing Multiplayer Sample over vanilla O3DE that I think only happens when users see the server start up the first time entering game mode.
Maybe once the initial cool factor runs out, I think ppl will want to know how to turn it off. This will be a much more visible option once we start showing network play options from the editor gui, so maybe hide it for now, but then turn it back on by default once the highly visible option is there to turn it off.

cgalvan (approved):

PR #147, opened 2022-05-12
WIP: Updating MultiplayerSample to use the Normal Spawnable API

AMZN-Gene:
Updating MP Sample project to use the normal spawnable api when creating players instead of networking CreateEntitiesImmediate

Tested by running the BaseSample level and making sure no errors/warnings occur when the player is loaded.

WIP: Still need to handle player disconnects.

### DIFF
--- a/Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
+++ b/Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
@@ -22,6 +22,7 @@
 #include <Multiplayer/Components/NetBindComponent.h>
 #include <Multiplayer/ConnectionData/IConnectionData.h>
 #include <Multiplayer/ReplicationWindows/IReplicationWindow.h>
+#include "AzCore/Component/TransformBus.h"
 
 namespace MultiplayerSample
 {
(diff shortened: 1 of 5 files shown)


Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
```diff
@@ -108,32 +109,33 @@ namespace MultiplayerSample
...
+            if (AZ::TransformInterface* transform = e->GetTransform())
+            {
+                transform->SetWorldTM(entityParams.second);
+            }
+        };
 
-        Multiplayer::NetworkEntityHandle controlledEntity;
-        if (!entityList.empty())
+        auto ticket = AZStd::make_unique<AzFramework::EntitySpawnTicket>(entityParams.first);
+        if (ticket->IsValid())
```
lmbr-pip, line 131: Does this need a log message or similar if the ticket is not valid? Is it ok to silently not spawn?

Gem/Code/Source/Spawners/IPlayerSpawner.h
```diff
@@ -34,7 +34,7 @@ namespace MultiplayerSample
         virtual ~IPlayerSpawner() = default;
 
         virtual bool RegisterPlayerSpawner(NetworkPlayerSpawnerComponent* spawner) = 0;
-        virtual AZStd::pair<Multiplayer::PrefabEntityId, AZ::Transform> GetNextPlayerSpawn() = 0;
+        virtual AZStd::pair<AZ::Data::Asset<AzFramework::Spawnable>, AZ::Transform> GetNextPlayerSpawn() = 0;
```
lmbr-pip, line 37: nit: feels like a using statement to set up a type here may help keep code more readable?

kberg-amzn (commented):
On just a cursory examination, this looks like roughly a wash in terms of lines of code added/removed and in fact maybe a slight in complexity of code since we're now dealing with tickets and assets and callbacks where before we weren't. Just so I can understand risk, what are the motivations and tradeoffs (pros/cons) of this change?

AMZN-Gene:
> On just a cursory examination, this looks like roughly a wash in terms of lines of code added/removed and in fact maybe a slight increase in complexity of code since we're now dealing with tickets and assets and callbacks where before we weren't. Just so I can understand risk, what are the motivations and tradeoffs (pros/cons) of this change?

This is a move to use proper spawnable API, and thus, not having to use manual serializeContext->Clone() which network entities currently do. The unfortunate side-effect being that normal spawnables are async and thus use callbacks :c
Also, this is more phase-2 edits... I'm going to close this, because I want for attention on phase-1. (see https://github.com/o3de/o3de/pull/9548/files).

Issue #157, opened 2022-06-27
MPSample Spawning Perf Test No Longer Working


AMZN-Gene:
**Describe the bug**
Levels/Player_SpawningPerfTest no longer works.

**Steps to reproduce**
1. Open Levels/Player_SpawningPerfTest
2. Press CTRL+G 

**Expected behavior**
Lots of golden boxes spawn

**Actual behavior**
Nothing occurs

**Found in Branch**
Development

PR #158, opened 2022-06-28
SpawningPerfTest Fix: Updated to Use Player Spawner

AMZN-Gene:
Fixing SpawningPerfTest by adding the new player spawner component to the SpawningPerfTest. 
Also applying random movements to the test boxes instead of using PhysX because PhysX had a performance bug.

Tested by opening SpawningPerfTest in Editor and pressing CTRL+G and now noticing that the boxes spawn and move around

Fixes GHI #157 

### DIFF
--- a/Gem/Code/Source/AutoGen/NetworkRandomTranslateComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/NetworkRandomTranslateComponent.AutoComponent.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+
+<Component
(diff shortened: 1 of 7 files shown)


AMZN-Olex (approved):

AMZN-puvvadar:
This should have been fixed with https://github.com/o3de/o3de-multiplayersample/pull/153. What caused this to resurface?

AMZN-Gene:
> This was previously fixed with #153. What caused this to resurface?

You’re right! Spawning is fixed in development latest. Thanks!

AMZN-puvvadar (approved):

PR #167, opened 2022-07-12
Remove FlyCameraComponent and update SimplePlayerCamera interpolation

AMZN-puvvadar:
This removes the FlyCameraComponent from SampleBase's camera. It fights with NetworkSimplePlayerCameraComponent on position causing a lot of jittering on the camera's position. This change also updates NetworkSimplePlayerCameraComponent to use blend factor to interpolate between current and previous aim angles much like NetworkTransformComponent.

### DIFF
--- a/Levels/SampleBase/SampleBase.prefab
+++ b/Levels/SampleBase/SampleBase.prefab
@@ -1925,13 +1925,6 @@
                     "$type": "EditorInspectorComponent",
                     "Id": 154105298091518109
                 },
-                "Component_[15452140819580244001]": {
-                    "$type": "GenericComponentWrapper",
-                    "Id": 15452140819580244001,
-                    "m_template": {
-                        "$type": "FlyCameraInputComponent"
-                    }
-                },
                 "Component_[17443734220531699641]": {
                     "$type": "EditorOnlyEntityComponent",
                     "Id": 17443734220531699641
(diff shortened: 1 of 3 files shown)


Levels/SampleBase/SampleBase.prefab
```diff
@@ -1925,13 +1925,6 @@
                     "$type": "EditorInspectorComponent",
                     "Id": 154105298091518109
                 },
-                "Component_[15452140819580244001]": {
-                    "$type": "GenericComponentWrapper",
-                    "Id": 15452140819580244001,
-                    "m_template": {
-                        "$type": "FlyCameraInputComponent"
```
santorac, line 1932: The FlyCameraInputComponent is what allows the user to navigate the world in the server window. Is there a way we can keep this component but only activate it on the server?

AMZN-puvvadar: That's interesting. What if we did an entirely separate camera entity for the server? @AMZN-Gene might have some thoughts here.

santorac: Done: https://github.com/o3de/o3de/issues/10633

AMZN-Gene (approved):

lumberyard-employee-dm (approved):

PR #170, opened 2022-07-27
Misc Multiplayer Sample CFG Cleanup

AMZN-Gene:
Misc CFG cleanup to help prevent workflow-6 (multiplayer) confusion. 
1) Removed unused autoexec.cfg; we can bring this back as needed for auto-launching a starter level for releases, but it wasn't being used and still using old cry map commands. 
2) Added Window's cmd files to partner up with the server/client cfg so users know those cfg's aren't automatically loaded and instead can/need to be used in a command-line.

Partially resolves https://github.com/o3de/o3de/issues/9630

**Testing**
Tested by opening MultiplayerSample editor and making ensuring all the cvars were as expected


AMZN-puvvadar:
Any concerns around relying on profile builds existing specifically?

AMZN-Gene:
> Any concerns around relying on profile builds existing specifically?

I’m okay with relying on profile builds. I use cmd’s like this myself during development and it’s an example for devs to quickly see how to use cfg’s. Alternatively, we could build out a more complicated/robust script that first searches for release builds, but I think the simplicity is an advantage for educational purposes (and it’ll be what ppl use most the time).

launch_client.cmd
```diff
@@ -0,0 +1 @@
+.\build\windows\bin\profile\MultiplayerSample.GameLauncher.exe --console-command-file=launch_client.cfg
```
AMZN-Olex, line 1: Do we need to update VS debug settings as well?

allisaurus (commented):
Love the idea of adding helper cmd scripts, just don't forget to [update the README](https://github.com/o3de/o3de-multiplayersample#running-the-server) with 1) the new .cfg file names, or 2) example usages

launch_client.cmd
AMZN-Gene: good catch, fixed!

AMZN-Gene:
> Love the idea of adding helper cmd scripts, just don't forget to [update the README](https://github.com/o3de/o3de-multiplayersample#running-the-server) with 1) the new .cfg file names, or 2) example usages

Fixed!

README.md
```diff
@@ -83,28 +83,28 @@ This option will output all the project and engine binaries in the engine's buil
...
-
 #### Running the Server
 
 A server can be run as follows
 
 ```shell
-MultiplayerSample.ServerLauncher.exe --console-command-file=server.cfg 
+MultiplayerSample.ServerLauncher.exe --console-command-file=launch_server.cfg 
 ```
+Notice the launch_server.cfg is passed into the commandline. Any file passed into the console-command-file argument will used when starting up the application.
```
allisaurus, line 106: nit: "will _be_ used..."

allisaurus (approved):

README.md
AMZN-Gene: thank you! fixed!

AMZN-Olex (approved):

Issue #171, opened 2022-07-27
Demo Level Placards Don't Appear on Server


AMZN-Gene:
**Describe the bug**
The demo level containing various networked and non-networked entity hierarchies contains placards that only appear in editor.

**Steps to reproduce**
1. Launcher ServerLauncher.exe
2. Open SampleBase level to view the demo objects

**Expected behavior**
The placards should appear on server showing what everything is trying to demostrate

**Actual behavior**
The placards dont appear

**Assets required**
Provide sample assets needed to reproduce the issue.

**Screenshots/Video**
![image](https://user-images.githubusercontent.com/32776221/181154901-9ac128e9-6cc5-448e-a4e3-88c011c3f54b.png)

**Found in Branch**
development July 26 2022

PR #173, opened 2022-07-27
Remove LyShine Dependency

AMZN-Gene:
No more placard component so removeLyShine dependency.
Tested by compiling.
Last step for #171 


AMZN-Olex:
We would still need UI for HUD, though, right?

AMZN-Olex (changes requested):
We still need to LyShine for HUD work, right?

lmbr-pip:
No longer needed

PR #176, opened 2022-07-28
ClientServer Fix

AMZN-Gene:
Adding entity preactivation callback to network spawners in order for multiplayer system to edit the network entities before it's activated (used to turn on autonomous role)
This change depends on an o3de change: https://github.com/o3de/o3de/pull/10951

### DIFF
--- a/Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
+++ b/Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
@@ -108,43 +108,7 @@ namespace MultiplayerSample
         return AZ::TICK_PLACEMENT + 2;
     }
 
-    void MultiplayerSampleSystemComponent::EnableAutonomousControl(Multiplayer::NetworkEntityHandle entityHandle, AzNetworking::ConnectionId connectionId)
-    {
-        if (!entityHandle.Exists())
-        {
-            AZLOG_WARN("Attempting to enable autonomous control for an invalid entity");
-            return;
-        }
-
-        entityHandle.GetNetBindComponent()->SetOwningConnectionId(connectionId);
-        if (connectionId == InvalidConnectionId)
-        {
-            entityHandle.GetNetBindComponent()->SetAllowAutonomy(true);
-        }
-
-        /*auto* hierarchyComponent = entityHandle.FindComponent<NetworkHierarchyRootComponent>();
-        if (hierarchyComponent != nullptr)
-        {
-            for (AZ::Entity* subEntity : hierarchyComponent->GetHierarchicalEntities())
-            {
-                NetworkEntityHandle subEntityHandle = NetworkEntityHandle(subEntity);
-                NetBindComponent* subEntityNetBindComponent = subEntityHandle.GetNetBindComponent();
-
-                if (subEntityNetBindComponent != nullptr)
-                {
-                    subEntityNetBindComponent->SetOwningConnectionId(connectionId);
-                    if (connectionId == InvalidConnectionId)
-                    {
-                        subEntityNetBindComponent->SetAllowAutonomy(true);
-                    }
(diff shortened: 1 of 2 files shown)


Gem/Code/Source/MultiplayerSampleSystemComponent.h
```diff
@@ -57,7 +57,7 @@ namespace MultiplayerSample
 
         ////////////////////////////////////////////////////////////////////////
         // IMultiplayerSpawner overrides
-        Multiplayer::NetworkEntityHandle OnPlayerJoin(uint64_t userId, const Multiplayer::MultiplayerAgentDatum& agentDatum) override;
+        Multiplayer::NetworkEntityHandle OnPlayerJoin(uint64_t userId, const Multiplayer::MultiplayerAgentDatum& agentDatum, EntityPreActivationCallback preActivationCallback) override;
```
AMZN-Olex, line 60: Is the api change for OnPlayerJoin already in or another PR that this one depends on?

AMZN-Olex (approved):

Gem/Code/Source/MultiplayerSampleSystemComponent.h
AMZN-puvvadar: This looks dependent on https://github.com/o3de/o3de/pull/10951. @AMZN-Gene can you mention that in the description?

Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
```diff
@@ -108,15 +108,16 @@ namespace MultiplayerSample
         return AZ::TICK_PLACEMENT + 2;
     }
 
-    Multiplayer::NetworkEntityHandle MultiplayerSampleSystemComponent::OnPlayerJoin(
-        [[maybe_unused]] uint64_t userId, [[maybe_unused]] const Multiplayer::MultiplayerAgentDatum& agentDatum)
+    Multiplayer::NetworkEntityHandle MultiplayerSampleSystemComponent::OnPlayerJoin([[maybe_unused]] uint64_t userId, [[maybe_unused]] const Multiplayer::MultiplayerAgentDatum& agentDatum, EntityPreActivationCallback preActivationCallback)
     {
         AZStd::pair<Multiplayer::PrefabEntityId, AZ::Transform> entityParams = AZ::Interface<IPlayerSpawner>::Get()->GetNextPlayerSpawn();
 
         Multiplayer::INetworkEntityManager::EntityList entityList =
```
AMZN-puvvadar, line 115: This feels like this could be collapsed in MultiplayerSystemComponent rather than a callback considering the callback is calling EnableAutotnomousControl. Could CreateEntitiesImmediate be handling the EnableAutonomousControl? Alternatively, could there be an IMultiplayer call added that handles EnableAutonomousControl?

Gem/Code/Source/MultiplayerSampleSystemComponent.h
AMZN-Gene: thank you!

Gem/Code/Source/MultiplayerSampleSystemComponent.cpp
AMZN-Gene: Cleaned this up, let me know what you think! We are really bound to having autonomy enabled _before_ entity activation. I ended up moving the enabling of autonomous control into NetworkEntityManager before the entities are activated; also added a helper method "CreateAutonomousPlayerImmediate" (a nicer/simpler method for IMultiplayerSpawner to call with less parameters). This way, each project's MultiplayerSpawner could go to town modifying the entities before passing it back to MultiplayerSystemComponent, and still know everything will just work.

AMZN-puvvadar (approved):

PR #194, opened 2022-09-15
Safety Check Round Robin Spawner Index

### LINKED ISSUE #193: Round Robin Spawner Bad Index
**Describe the bug**
If the server changes levels the spawn index can be out-of-bounds and cause a crash.

**Steps to reproduce**
1. Open a server and start hosting a level with more than 1 spawner
2. Have a client join (this causes round robin m_spawnIndex to be 1)
3. Disconnect from multiplayer simulation
4. Host a new level this time with only one spawner in it
5. Have client join

**Expected behavior**
Game doesn't crash, and player joins at the spawner location

**Actual behavior**
Crash. m_spawnIndex >= m_spawners.size()

**Found in Branch**
development latest

AMZN-Gene:
Make sure round robin spawn index is reset after unregistering.
Also added safety net to check for out-of-bounds spawn index if needed

Fixes #193 

### DIFF
--- a/Gem/Code/Source/Spawners/RoundRobinSpawner.cpp
+++ b/Gem/Code/Source/Spawners/RoundRobinSpawner.cpp
@@ -32,7 +32,7 @@ namespace MultiplayerSample
 
         if (m_spawnIndex >= m_spawners.size())
         {
-            AZLOG_WARN("RoundRobinSpawner has an out-of-bounds spawner index. Resetting spawn index to 0.")
+            AZLOG_WARN("RoundRobinSpawner has an out-of-bounds spawner index. Resetting spawn index to 0. Did you forget to call UnregisterPlayerSpawner?")
             m_spawnIndex = 0;
         }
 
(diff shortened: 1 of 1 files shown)


Gem/Code/Source/Spawners/RoundRobinSpawner.cpp
```diff
@@ -30,6 +30,12 @@ namespace MultiplayerSample
             return AZStd::make_pair<Multiplayer::PrefabEntityId, AZ::Transform>(Multiplayer::PrefabEntityId(), AZ::Transform::CreateIdentity());
         }
 
+        if (m_spawnIndex >= m_spawners.size())
+        {
+            AZLOG_WARN("RoundRobinSpawner has an out-of-bounds spawner index. Resetting spawn index to 0.")
```
lmbr-pip, line 35: Would question if this a warning ie we fixed it and do not appear to expect user to take action. Is it worth having log message here?

AMZN-Gene: This will never get hit if ppl use our SpawnerComponent. It could happen if some other spawner the user wrote didn't unregister itself. I'd like users to know if it happens, because if it does something is off with their logic.

AMZN-puvvadar (approved):

lmbr-pip (approved):

PR #200, opened 2022-09-30
MultiplayerSample CompileDef based Client/Server split

AMZN-puvvadar:
This change relates to: https://github.com/o3de/o3de/pull/12344 and implements the corresponding changes to MultiplayerSample.


Gem/Code/Source/Components/NetworkStressTestComponent.cpp
```diff
@@ -38,7 +38,9 @@ namespace MultiplayerSample
 
     NetworkStressTestComponentController::NetworkStressTestComponentController(NetworkStressTestComponent& owner)
         : NetworkStressTestComponentControllerBase(owner)
+#if AZ_TRAIT_SERVER_ENABLED
         , m_autoSpawnTimer([this]() { HandleSpawnAiEntity(); }, AZ::Name("StressTestSpawner Event"))
+#endif
```
AMZN-puvvadar, line 43: This pattern is something I'm not particularly fond of. It occurs because HandleSpawnAiEntity is not declared on the Client. An alternative is to declare it on the client but define it as empty. The tradeoff here is that the Client has a concept of HandleSpawnAiEntity in exchange for simplifying some of the macro usage around it.

Gem/Code/Source/Components/NetworkHealthComponent.cpp
```diff
@@ -24,10 +24,12 @@ namespace MultiplayerSample
         ;
     }
 
+#if AZ_TRAIT_SERVER_ENABLED
```
AMZN-Olex, line 27: AZ_TRAIT_SERVER should be shorter without losing meaning, imho, as you suggested.

Gem/Code/Source/AutoGen/NetworkRandomComponent.AutoComponent.xml
```diff
@@ -8,6 +8,6 @@
     OverrideInclude="Source/Components/NetworkRandomComponent.h"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
-    <NetworkProperty Type="uint64_t" Init="0" Name="Seed" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="false" IsPredictable="false" ExposeToEditor="false" ExposeToScript="false" GenerateEventBindings="false" Description="The RNG seed" />
+    <NetworkProperty Type="uint64_t" Init="0" Name="Seed" ReplicateFrom="Authority" ReplicateTo="Client" Container="Object" IsPublic="true" IsRewindable="false" IsPredictable="true" ExposeToEditor="false" ExposeToScript="false" GenerateEventBindings="false" Description="The RNG seed" />
```
AMZN-Gene, line 11: IsPredicable = true. That gives autonomous write access right? What if the player roles the next number and then does something that affects the mp state before the server also got a chance to role the same number? Does it also need to be rewindable?

AMZN-Gene (approved):

AMZN-Olex (approved):

Gem/Code/Source/AutoGen/NetworkRandomComponent.AutoComponent.xml
AMZN-puvvadar: Yup, we're hoping that the two stay in sync via determinism of the rolls and that desyncs are corrected at the level of the affected value rather than against the seed.

Gem/Code/multiplayersample_files.cmake
```diff
@@ -11,18 +11,6 @@ set(FILES
     # Scripting samples
     Source/AutoGen/ScriptingPlayerMovementComponent.AutoComponent.xml
```
AMZN-Gene, line 12: Can this move to _autogen_files.cmake as well? It would be good to have all the XMLs in one place.

Gem/Code/multiplayersample_autogen_files.cmake
```diff
@@ -11,4 +11,16 @@ set(FILES
     ${LY_ROOT_FOLDER}/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Source.jinja
     ${LY_ROOT_FOLDER}/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponentTypes_Header.jinja
     ${LY_ROOT_FOLDER}/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponentTypes_Source.jinja
+    Source/AutoGen/NetworkAiComponent.AutoComponent.xml
```
AMZN-Gene, line 14: Can we glob anything ending with AutoComponent.xml? Provide a comment showing an example of how to subtract a file if anyone ever needs to.

AMZN-Gene (approved):

Gem/Code/multiplayersample_files.cmake
AMZN-puvvadar: Might as well now that I've removed the AutoGen target.

Gem/Code/multiplayersample_autogen_files.cmake
AMZN-puvvadar: This function exactly as it did previously. I just moved AutoGen jinjas and xmls into a separate cmake.

PR #203, opened 2022-10-12
Setting up NPC Mode

lmbr-pip:
Seems like I need a property on the player spawner.

Looking for advice how to do this dynamically


lmbr-pip:
Branch against wrong repro

PR #211, opened 2023-01-12
Add Client and Server Module declarations for MPS

AMZN-puvvadar:
This is a companion change to https://github.com/o3de/o3de/pull/14031 to fix RELEASE builds


AMZN-Gene:
AutomatedTesting will need this update too?

AMZN-Gene (approved):

AMZN-puvvadar:
> AutomatedTesting will need this update too?

It won't actually. This is only required for projects that declare different Client, Server and Unified modules like MPS and Multiplayer Gem. Projects that alias all three to the same thing will be fine.

cgalvan (approved):

PR #214, opened 2023-01-23
Add build config for MPS client/server AR

amzn-changml:
Adds project specific jobs to build MPS client and server targets (`MultiplayerSample.GameLauncher` and  `MultiplayerSample.ServerLauncher`). This will overlay only the jobs defined in `scripts/Build` into the AR pipeline. The default AR jobs will need to be re-defined here if desired.

Tested on a sandbox Jenkins with a new Jenkinsfile configuration in a o3de engine branch.

Signed-off-by: Mike Chang <changml@amazon.com>


byrcolin (approved):

scripts/build/Platform/Linux/build_config.json
```diff
@@ -0,0 +1,36 @@
+{
+  "clean": {
+    "TAGS": [],
+    "COMMAND": "clean_linux.sh",
+    "PARAMETERS": {
+      "OUTPUT_DIRECTORY": "build",
+      "CMAKE_LY_PROJECTS": "AutomatedTesting"
```
AMZN-puvvadar, line 7: Does the project matter here at all? My only concern would be old MPS codegen artifacts not getting flushed.

amzn-changml: Not really, it was just the default config in the o3de AR. AR will still use `AutomatedTesting` if it's not defined. The codegen artifacts will get flushed once we get the Jenkinsfile changes in, since it will start with a clean build.

PR #233, opened 2023-03-01
Removed explicit template arguments from `AZStd::make_pair`

lumberyard-employee-dm:
The make_pair function template is meant for template argument deduction.
If the template arguments are to be specified, then the `AZStd::pair` template should be used directly.

Other changes: Added the CMakeUserPresets.json to the .gitignore

# How was this PR Tested?
Built the Multiplayer Sample project on Windows using non-unity files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ _savebackup/
 [Bb]uild/
 [Cc]ache/
 [Uu]ser/
+CMakeUserPresets.json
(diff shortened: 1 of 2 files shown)


mbalfour-amzn (approved):

amzn-phist (approved):
Thank you!

cgalvan (approved):

PR #242, opened 2023-03-16
Replacing the non-portable reference to the Multiplayer Gem location with a portable reference

### LINKED ISSUE #44: MultiplayerSample Project needs to update its reference to the Multiplayer Gem jinja templates
The Multiplayer Sample contains a reference to jinja template files within the Multiplayer Gem that are not available in an SDK layout build.
 The [multiplayersample_autogen_files.cmake](https://github.com/o3de/o3de-multiplayersample/blob/development/Gem/Code/multiplayersample_autogen_files.cmake) is referencing jinja template files from the Engine Source, which doesn't get copied to the Install SDK layout due to being in a private directory.

Furthermore that cmake file hardcodes the directory of the Multiplayer Gem location to be `${LY_ROOT_FOLDER}/Gems/Multiplayer` which isn't portable and would cause the MultiplayerSample project to fail to configure if the Multiplayer Gem was relocated outside of the Engine Source Repo or the folder was renamed.

Instead what should be done is to access the source directory of the Multiplayer Gem, that way the Multiplayer Gem can be relocated without breaking the Multiplayer Sample project
Running the `get_property(multiplayer_gem_source_dir TARGET Multiplayer PROPERTY SOURCE_DIR)` command can be used to lookup the source directory without using a generator expression

lumberyard-employee-dm:
If the Multiplayer Gem location is ever moved or renamed it would break the autogen references in the project

fixes #44

## How was this PR tested?

Configured and built the Multiplayer Sample project successfully

### DIFF
--- a/Gem/Code/multiplayersample_autogen_files.cmake
+++ b/Gem/Code/multiplayersample_autogen_files.cmake
@@ -5,12 +5,14 @@
 #
 #
 
+get_property(multiplayer_gem_root GLOBAL PROPERTY "@GEMROOT:Multiplayer@")
+
 set(FILES
-    ${LY_ROOT_FOLDER}/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Common.jinja
-    ${LY_ROOT_FOLDER}/Gems/Multiplayer/Code/Include/Multiplayer/AutoGen/AutoComponent_Header.jinja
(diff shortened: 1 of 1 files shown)


mbalfour-amzn (approved):

lmbr-pip (approved):

lumberyard-employee-dm:
> Assume this needs to go stabilization branch?

It shouldn't need to go into stabilization.

Even without the change, as long as the Multiplayer Gem isn't relocated in O3DE repo between the upcoming release and the next one after, it should cause any problems

PR #244, opened 2023-03-17
Fixed mouse aiming

AMZN-Olex:
Main issue fixed:
- dropping some of the input data that were ignoring all the last press and hold series of inputs between CreateInput calls

### DIFF
--- a/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
+++ b/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
@@ -26,8 +26,8 @@
 namespace MultiplayerSample
 {
     AZ_CVAR(float, cl_WasdStickAccel, 5.0f, nullptr, AZ::ConsoleFunctorFlags::Null, "The linear acceleration to apply to WASD inputs to simulate analog stick controls");
-    AZ_CVAR(float, cl_AimStickScaleZ, 0.05f, nullptr, AZ::ConsoleFunctorFlags::Null, "The scaling to apply to aim and view adjustments");
-    AZ_CVAR(float, cl_AimStickScaleX, 0.025f, nullptr, AZ::ConsoleFunctorFlags::Null, "The scaling to apply to aim and view adjustments");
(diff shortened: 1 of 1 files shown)


mbalfour-amzn (approved):

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
```diff
@@ -127,8 +139,8 @@ namespace MultiplayerSample
 
         // View Axis are clamped and brought into the -1,1 range for transport across the network.
         // These are set regardless of whether or not any other player actions are allowed.
-        playerInput->m_viewYaw = MouseAxis(AZStd::clamp<float>(m_viewYaw, -cl_MaxMouseDelta, cl_MaxMouseDelta) / cl_MaxMouseDelta);
-        playerInput->m_viewPitch = MouseAxis(AZStd::clamp<float>(m_viewPitch, -cl_MaxMouseDelta, cl_MaxMouseDelta) / cl_MaxMouseDelta);
+        playerInput->m_viewYaw = MouseAxis(0.5f * AZStd::clamp<float>(m_viewYaw, -cl_MaxMouseDelta, cl_MaxMouseDelta) / cl_MaxMouseDelta);
```
AMZN-Gene, line 142: Should this send value between <-1/2, 1/2> or <-1, 1>? Looking at the comment above it seems like MouseAxis wants <-1,1>

AMZN-Gene (approved):

AMZN-Gene:
https://github.com/o3de/o3de-extras/issues/73

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
kberg-amzn: I think this will effectively half the mouse movement speed in mps, which should help aiming but perhaps be a but annoying when traversing around the world? We might eventually want a user-configurable scalar in the settings registry for mouse acceleration?

kberg-amzn (approved):

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
AMZN-Olex: You are correct! Thanks!

PR #250, opened 2023-03-20
Automatically Show Score Between Rounds

AMZN-Gene:
- Automatically show score screen when round is over so that players can quickly see their score.
- Automatically hide the screen 3 seconds before the next round starts.
- Update HUD so score screen renders on top of round rest countdown

Tested by playing NewStarBase in Editor and seeing the score screen pop and tear-down between rounds.
No error window popup when opening NewStarBase.

### DIFF
--- a/Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
+++ b/Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
@@ -98,19 +98,19 @@ namespace MultiplayerSample
 
     void UiMatchPlayerCoinCountsComponent::EnableUI(bool enable)
     {
+        UiElementBus::Event(m_rootElementId, &UiElementBus::Events::SetIsEnabled, enable);
+
         if (enable)
         {
-            UiElementBus::Event(m_rootElementId, &UiElementBus::Events::SetIsEnabled, true);
             if (!m_onPlayerScoreChanged.IsConnected())
             {
                 AZ::Interface<MatchPlayerCoinsComponent>::Get()->CoinsPerPlayerAddEvent(m_onPlayerScoreChanged);
             }
             UpdatePlayerScoreUI();
         }
-        else if (!enable)
+        else
         {
             m_onPlayerScoreChanged.Disconnect();
-            UiElementBus::Event(m_rootElementId, &UiElementBus::Events::SetIsEnabled, false);
         }
     }
 
--- a/Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.h
+++ b/Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.h
@@ -19,6 +19,8 @@ namespace MultiplayerSample
         , public StartingPointInput::InputEventNotificationBus::MultiHandler
     {
     public:
+        static constexpr float SecondsBeforeNewRoundToHideUI = 3.0f;
+
         AZ_COMPONENT(UiMatchPlayerCoinCountsComponent, "{529b9b3b-bea2-4120-9089-c4451438e4c0}");
 
         static void Reflect(AZ::ReflectContext* context);
@@ -63,7 +65,7 @@ namespace MultiplayerSample
         // Automatically close the player score menu when the round is about to start
         AZ::Event<RoundTimeSec>::Handler m_restTimerHandler{ [this](RoundTimeSec secondsRemaining)
         {
-            if (secondsRemaining <= 3.0f)
+            if (secondsRemaining <= SecondsBeforeNewRoundToHideUI)
             {
                 EnableUI(false);
                 m_restTimerHandler.Disconnect();


mbalfour-amzn (approved):

Gem/Code/Source/Components/UI/HUDComponent.cpp
```diff
@@ -25,6 +25,8 @@ namespace MultiplayerSample
     void HUDComponent::Deactivate()
     {
         m_waitForActiveNetworkMatchComponent.RemoveFromQueue();
+        m_roundNumberHandler.Disconnect();
```
lmbr-pip, line 28: nit: where do these attach? Assume its because this is some type of base component?

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
```diff
@@ -26,12 +26,15 @@ namespace MultiplayerSample
...
     {
+        m_waitForActiveNetworkMatchComponent.Enqueue(AZ::TimeMs{ 1000 }, true);
         StartingPointInput::InputEventNotificationBus::MultiHandler::BusConnect(ShowPlayerCoinCountsEventId);
     }
 
     void UiMatchPlayerCoinCountsComponent::Deactivate()
     {
         StartingPointInput::InputEventNotificationBus::MultiHandler::BusDisconnect();
+        m_waitForActiveNetworkMatchComponent.RemoveFromQueue();
+        m_roundTimerHandler.Disconnect();
```
lmbr-pip, line 37: Does this need to detach from m_roundTimerHandler()? Or can the code the base class deactivate instead?

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
```diff
@@ -93,6 +96,24 @@ namespace MultiplayerSample
         }
     }
 
+    void UiMatchPlayerCoinCountsComponent::EnableUI(bool enable)
+    {
+        if (enable)
+        {
+            UiElementBus::Event(m_rootElementId, &UiElementBus::Events::SetIsEnabled, true);
```
lmbr-pip, line 103: nit: can pass `enable` rather than `true`

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
```diff
@@ -93,6 +96,24 @@ namespace MultiplayerSample
...
+            if (!m_onPlayerScoreChanged.IsConnected())
+            {
+                AZ::Interface<MatchPlayerCoinsComponent>::Get()->CoinsPerPlayerAddEvent(m_onPlayerScoreChanged);
+            }
+            UpdatePlayerScoreUI();
+        }
+        else if (!enable)
```
AMZN-Olex, line 110: Shouldn't this be just ```else {```

AMZN-Olex (approved):

Gem/Code/Source/Components/UI/HUDComponent.cpp
AMZN-Gene: It connects once ActiveNetworkMatchComponent. UI lives on a different entity then the network stuff, so we cant use the normal component dependency to activate UI once the network match component is activated. 
An alternative could be to broadcast a notification when the NetworkMatchComponent turns on, and then listen for that message in UI.
See where m_waitForActiveNetworkMatchComponent is defined in the header.

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
AMZN-Gene: This score screen doesn't listen for round changes, so it doesn't have a m_roundNumberHandler.
The reason all these UI screens have a m_waitForActiveNetworkMatchComponent is because they live on a separate entity so cant use the built in component dependency system. Alternatively we could implement a Network Match Component notification when it's activated, and listen for it.

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
AMZN-Gene: Thanks! Fixed

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.cpp
AMZN-Gene: fixed!

Gem/Code/Source/Components/UI/UiMatchPlayerCoinCountsComponent.h
```diff
@@ -45,5 +48,40 @@ namespace MultiplayerSample
...
+                    networkMatchComponent->AddRoundTimeRemainingEventHandler(m_roundTimerHandler);
+                    m_waitForActiveNetworkMatchComponent.RemoveFromQueue();
+                }
+            }, AZ::Name("GameOverUI Wait For Active NetworkMatchComponent"));
+
+        // Listen for rest time between rounds coming to an end
+        // Automatically close the player score menu when the round is about to start
+        AZ::Event<RoundTimeSec>::Handler m_restTimerHandler{ [this](RoundTimeSec secondsRemaining)
+        {
+            if (secondsRemaining <= 3.0f)
```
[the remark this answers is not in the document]
AMZN-Gene: fixed!

PR #253, opened 2023-03-21
Temp logging to help diagnose movement "disconnect"

mbalfour-amzn:
Currently, MPS has a bug where a player stops getting movement updates on the server. They can move around locally, but the replicated player stops moving. This logging is a temporary addition to help diagnose whether the input is getting discarded in ProcessInput() or if it's getting discarded somewhere else somehow. It will be removed again once we verify if the problem is occurring in ProcessInput() or somewhere else.

### DIFF
--- a/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
+++ b/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
@@ -200,6 +200,11 @@ namespace MultiplayerSample
         NetworkPlayerMovementComponentNetworkInput* playerInput = input.FindComponentInput<NetworkPlayerMovementComponentNetworkInput>();
         if (playerInput->m_resetCount != GetNetworkTransformComponentController()->GetResetCount())
         {
+            // TEMPORARY LOGGING TO HELP DIAGNOSE MOVEMENT DISCONNECTIONS.
+            // Specifically, sometimes a player will still be able to move around locally, but on the server and other connected clients,
+            // the player has stopped moving. They can no longer interact with the environment, but they are still connected and can see
+            // the other players moving around, still receive damage notifications,e tc.
+            // This logging will get removed after the root cause has been found and resolved.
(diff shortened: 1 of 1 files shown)


Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
```diff
@@ -200,6 +200,15 @@ namespace MultiplayerSample
         NetworkPlayerMovementComponentNetworkInput* playerInput = input.FindComponentInput<NetworkPlayerMovementComponentNetworkInput>();
         if (playerInput->m_resetCount != GetNetworkTransformComponentController()->GetResetCount())
         {
+            // TEMPORARY LOGGING TO HELP DIAGNOSE MOVEMENT DISCONNECTIONS.
+            // Specifically, sometimes a player will still be able to move around locally, but on the server and other connected clients,
+            // the player has stopped moving. They can no longer interact with the environment, but they are still connected and can see
+            // the other players moving around, still receive damage notifications,e tc.
+            // This logging will get removed after the root cause has been found and resolved.
+            AZLOG_INFO("netEntityId=%u: Different reset count, discarding player input. Input / local reset=%u / %u, clientInputId=%u, hostFrame=%u",
```
AMZN-Olex, line 208: When his happens do you see a non-stop corrections being sent to that client?

AMZN-Olex (approved):

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
mbalfour-amzn: I don't _think_ so, but I'm not 100% sure, since the only log I have is from a pretty long play session with a lot of corrections in it to begin with. But since this is the only early-return in ProcessInput(), this will at least tell us if we're hitting that pathway when the problem occurs, or else if the problem is occurring somewhere else in the chain (like in LocalPrediction...cpp or something).

PR #256, opened 2023-03-21
Sprint cam tweaks

kberg-amzn:
Some minor tweaks to the sprint camera to make it less disruptive when running and gunning. This reduces the magnitude of the effect, while only enabling it if the player is actively moving forward. Transition time in and out of the effect is reduced to make it less distracting. Fixes a broken debug cvar in the weapons code that is used to halt simulation when diagnosing hit detection issues.

### DIFF
--- a/Gem/Code/Source/Weapons/BaseWeapon.cpp
+++ b/Gem/Code/Source/Weapons/BaseWeapon.cpp
@@ -125,7 +125,7 @@ namespace MultiplayerSample
         const bool result = MultiplayerSample::GatherEntities(m_weaponParams.m_gatherParams, eventData, m_gatheredNetEntityIds, outResults);
         if (gp_PauseOnWeaponGather && (outResults.size() > 0))
         {
-            AZ::Interface<AZ::IConsole>::Get()->PerformCommand("t_scale 0");
+            AZ::Interface<AZ::IConsole>::Get()->PerformCommand("t_simulationTickScale 0");
         }
         return result;
     }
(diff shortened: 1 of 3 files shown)


Gem/Code/Source/Weapons/BaseWeapon.cpp
```diff
@@ -125,7 +125,7 @@ namespace MultiplayerSample
         const bool result = MultiplayerSample::GatherEntities(m_weaponParams.m_gatherParams, eventData, m_gatheredNetEntityIds, outResults);
         if (gp_PauseOnWeaponGather && (outResults.size() > 0))
         {
-            AZ::Interface<AZ::IConsole>::Get()->PerformCommand("t_scale 0");
+            AZ::Interface<AZ::IConsole>::Get()->PerformCommand("t_simulationTickScale 0");
```
AMZN-Olex, line 128: @AMZN-Gene recently into an issue around this call. Do we need to revert this to 1 or 0, Gene?

mbalfour-amzn: Seems like this change should also be made on line 138? This is a bit of an odd feature to me, because it requires the user to understand how to get back out of the state they've ended up in. At a minimum, the cvar comment on line 15 really ought to mention that the user should set t_simulationTickScale back to 1 when they're done examining the state.

AMZN-Olex (approved):

HogJonny-AMZN (approved):

PR #257, opened 2023-03-22
321 bug fixes

mcphedar:
Fixes the teleport locations that were placed in a tree and adds a few more late game gem spawns to complicate the "balfour speed run"

### DIFF
--- a/Levels/NewStarbase/NewStarbase.prefab
+++ b/Levels/NewStarbase/NewStarbase.prefab
@@ -198,7 +198,8 @@
                     "$type": "EditorEntitySortComponent",
                     "Id": 7768532315355253369,
                     "Child Entity Order": [
-                        "Instance_[1542952924443793]/ContainerEntity"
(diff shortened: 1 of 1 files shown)


Levels/NewStarbase/NewStarbase.prefab
```diff
@@ -4383,8 +4376,9 @@
                     "Id": 8615741586168280576,
                     "m_template": {
                         "$type": "MultiplayerSample::HUDComponent",
-                        "RoundNumberId": 8,
-                        "RoundTimerId": 4
+                        "RoundNumberId": "",
+                        "RoundTimerId": "",
+                        "RoundSecondsRemaining": ""
```
mbalfour-amzn, line 4381: Is this change intentional? If so, what does it fix? If not, does the HUD still work?

mcphedar: No this was not something I changed and no, the HUD doesn't work. I know it worked in testing earlier, so i'm not sure what happened.

PR #262, opened 2023-03-22
Lowered the volume of foot steps and laser fire.

AMZN-Olex:
- Using correct gun fire and impact sounds
- Lowered the volume of gun fire
- Lowered the volume of foot steps
- Cleaned up old temporary sounds that shouldn't be used


lmbr-pip (approved):

mbalfour-amzn (approved):

Prefabs/Player.prefab
```diff
@@ -530,12 +530,13 @@
...
                                     },
-                                    "AudioTrigger": "Play_Sound_Gun_Firing"
+                                    "AudioTrigger": "play_sx_wpn_laserpistol_fire"
                                 },
                                 "ImpactFx": {
                                     "ParticleAsset": {
                                         "guid": "{1BB94CA2-0BA0-5EC6-BE46-9B1F51D66CB9}"
-                                    }
+                                    },
+                                    "AudioTrigger": "play_sx_wpn_laserpistol_impact"
```
kberg-amzn, line 539: This will double-play the sound.. While the impact and damage effects overlay nicely, we probably only want the actual sound effect on either the impact or the damage. I put it on the damage so there's more feedback about what the server did.

kberg-amzn (approved):
The sound effect really needs to be fixed. The rest of the changes I can't really review. We'll have to playtest these changes to know for sure if it's a quality improvement. A lot of the provided sounds unfortunately had a fair bit of static, or didn't que well with things visually.

PR #263, opened 2023-03-23
Attaching a gun to each player using a new C++ component in a multiplayer friendly way

AMZN-Olex:
- Adds AttachPlayerWeaponComponent
- Modified Player.prefab to add AttachPlayerWeaponComponent to attach laser_gun.prefab to the player's right hand
- Modified Player.prefab by deleting a child entity with a gun that sadly doesn't work in multiplayer configuration
- Here is an example a moment after firing a shot and the recoil lifting the gun a bit:
image
- While aiming:
image


shawstar (commented):
Looks Great!

Gem/Code/Source/Components/AttachPlayerWeaponComponent.h
```diff
@@ -0,0 +1,40 @@
...
+        void Activate() override;
+        void Deactivate() override;
+        //! }@
+
+    private:
+        AZ::Data::Asset<AzFramework::Spawnable> m_gunAsset;
+        AZStd::string m_boneToAttachTo;
+        AZ::Transform m_attachmentTransform;
+
+        AZStd::shared_ptr<AzFramework::EntitySpawnTicket> m_gunTicket;
```
lmbr-pip, line 38: Same with this m_weaponTicket or better yet m_weaponSpanTicket

Gem/Code/Source/Components/AttachPlayerWeaponComponent.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Math/Transform.h>
+#include <AzFramework/Spawnable/Spawnable.h>
+#include <AzFramework/Spawnable/SpawnableEntitiesInterface.h>
+
+namespace MultiplayerSample
+{
+    //! @class AttachPlayerWeaponComponent
```
lmbr-pip, line 17: Comment on public class! Nice.

PR #264, opened 2023-03-24
Client Automatically Opens Start Menu

AMZN-Gene:
- GameLauncher.exe setreg update to load the start menu level
- Converting existing ScriptCanvas start menu to C++
- Adding start up messaging when attempting to connect to host and when connection fails
- Disabling the host button for launchers that don't support hosting


kberg-amzn (approved):
Just out of curiosity, not asking to make the change at this point, but is the native component actually necessary? What is missing from our script-bind that prevents all of this from being implemented in script canvas?
Otherwise, awesome! Much needed changes for release.

Registry/settings.multiplayersample_gamelauncher.setreg
```diff
@@ -0,0 +1,11 @@
+{
+    "Amazon": {
+        "AzCore": {
+            "Runtime": {
+                "ConsoleCommands": {
+                    "loadlevel": "startmenu"
```
AMZN-Olex, line 6: Does this apply to Server Launcher?

AMZN-Gene: Nay, this PR only adds to __gamelauncher.setreg_

Suppose we could also add this to server launchers... I'd want to make sure loading a level in setreg for servers doesn't override editor-server does when it launches its server. Last I checked, setting the server setreg loadlevel _would_ load that lever, and then switch over to the original level the editor was attempting to test. But ideally we'd want to a way to suppress that setreg loadlevel when the editor runs CTRL+G.

AMZN-Gene:
> Just out of curiosity, not asking to make the change at this point, but is the native component actually necessary? What is missing from our script-bind that prevents all of this from being implemented in script canvas? Otherwise, awesome! Much needed changes for release.

Probably doable in script canvas, but I'd avoid writing anything one-off fire-and-forget type responses to events in script canvas. Large graphs aren't easy on the eyes, and sometimes graph assets break and are difficult to debug, like recently seen with reticle graph breaking during asset load inside lua.

AMZN-stankowi (approved):

AMZN-alexpete (approved):

Gem/Code/Source/Components/UI/UiStartMenuComponent.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        if (buttonEntityId == m_exitButtonUi)
+        {
+            console->PerformCommand("quit");
+        }
+        else if (buttonEntityId == m_hostButtonUi)
+        {
+            console->PerformCommand("host");
+            console->PerformCommand("loadlevel NewStarbase");
```
mbalfour-amzn, line 102: Really feels like the next level to load should at least be a parameter on the component.

AMZN-Gene: Originally had that as a variable in the UI, but felt like a variable I wouldn't want art to accidentally change/misspell. Maybe a class static constexpr would be more appropriate.

Gem/Code/Source/Components/UI/UiStartMenuComponent.h
```diff
@@ -0,0 +1,46 @@
...
+
+#pragma once
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Settings/SettingsRegistry.h>
+#include <Multiplayer/IMultiplayer.h>
+
+namespace MultiplayerSample
+{
+    class UiStartMenuComponent
```
lmbr-pip, line 17: nit: comment on all public methods please

lmbr-pip (commented):
From a UX POV, do we autofill the server and port if they are set by cvars?

AMZN-Gene:
> From a UX POV, do we autofill the server and port if they are set by cvars?

I would LOVE to! I couldn't seem to access a valid cvar on activate from this component. I _could_ add a few ticks and probably grab it, but I really need to debug why the cvars aren't available. The UI is activated well after all systems components are stood up.

Gem/Code/Source/Components/UI/UiStartMenuComponent.cpp
AMZN-Gene: fixed!

mbalfour-amzn: "Originally had that as a variable in the UI, but felt like a variable I wouldn't want art to accidentally change/misspell." - you can make it an Asset Reference that can only be set to a Spawnable type. This solves the misspelling problem and also gives the asset dependency system more information for creating bundles.

PR #265, opened 2023-03-24
Energy cannon

kberg-amzn:
Fixing up the energy cannon. Adding assets and effects, removing setreg and knockback, and adding component based configuration for designers to tune directly within the editor. Adds script-bind support for implementing effects like the jump-pad.

image

image

image

https://user-images.githubusercontent.com/82241079/228032309-b63fa617-5e13-48b2-b6b7-38d0ee291a6d.mp4


Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -5,120 +5,144 @@
...
+        AZ::Transform transform = AZ::Transform::CreateFromQuaternionAndTranslation(AZ::Quaternion::CreateIdentity(), location);
+        m_effect.TriggerEffect(transform);
+    }
+#endif
+
+
     EnergyBallComponentController::EnergyBallComponentController(EnergyBallComponent& parent)
         : EnergyBallComponentControllerBase(parent)
     {
+        ;
```
mbalfour-amzn, line 54: Nit: Unnecessary semicolon

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -5,120 +5,144 @@
...
-        LoadEnergyBallSettings();
-
-        AzPhysics::RigidBody* body = nullptr;
-        Physics::RigidBodyRequestBus::EventResult(body, GetEntityId(), &Physics::RigidBodyRequests::GetRigidBody);
-        if (body)
-        {
-            body->RegisterOnCollisionBeginHandler(m_collisionHandler);
-        }
+#if AZ_TRAIT_SERVER
+        m_collisionCheckEvent.Enqueue(AZ::TimeMs{ 10 }, true);
```
mbalfour-amzn, line 60: Why is 10 ms chosen here? It seems pretty arbitrary and disconnected from any other loop timing choices.

Gem/Code/Source/Components/Multiplayer/EnergyCannonComponent.cpp
```diff
@@ -15,6 +15,36 @@
...
+
+    void EnergyCannonComponent::OnActivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
+    {
+        m_effect = GetFiringEffect();
+        m_effect.Initialize();
+    }
+
+    void EnergyCannonComponent::OnDeactivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
+    {
+        ;
```
mbalfour-amzn, line 37: Nit: Unnecessary semicolon.

mbalfour-amzn (changes requested):

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
```diff
@@ -278,6 +278,19 @@ namespace MultiplayerSample
...
+    void NetworkPlayerMovementComponentController::HandleApplyImpulse([[maybe_unused]] AzNetworking::IConnection* connection, const AZ::Vector3& impulse)
+    {
+        const AZ::Vector3 newVelocity = GetSelfGeneratedVelocity() + impulse;
+        SetSelfGeneratedVelocity(newVelocity);
+    }
+
+    void NetworkPlayerMovementComponentController::HandleSetVelocity([[maybe_unused]] AzNetworking::IConnection* connection, const AZ::Vector3& velocity)
+    {
+        SetSelfGeneratedVelocity(velocity);
+    }
```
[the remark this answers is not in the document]
AMZN-alexpete: If the velocity/impulse isn't coming from the player lets use velocity from external sources so we can distinguish between player input velocity and stuff the player cannot control.

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
kberg-amzn: Correct, completely arbitrary. I want it to tick every frame with best effort, but it's not a super critical system and so it's okay if the other 0 ms events starve this for a frame or two.

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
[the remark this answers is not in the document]
kberg-amzn: When I tried setting the external velocity it appeared to be clobbered immediately and had no effect on actual player movement? I can try using the external velocity again however..

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
kberg-amzn: I usually put these just to remind myself that yeah, it's really an empty function and not a method that was stubbed and missed, but it seems to trip up enough people that I'll just remove them.

Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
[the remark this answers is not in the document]
kberg-amzn: After some experimentation, setting external velocity works but it requires a significantly larger impulse. I'll make the switch and post the updated PR shortly.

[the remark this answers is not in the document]
kberg-amzn: Done

Levels/NewStarbase/NewStarbase.prefab
```diff
@@ -4383,8 +4383,9 @@
                     "Id": 8615741586168280576,
                     "m_template": {
                         "$type": "MultiplayerSample::HUDComponent",
-                        "RoundNumberId": 8,
-                        "RoundTimerId": 4
+                        "RoundNumberId": "",
+                        "RoundTimerId": "",
+                        "RoundSecondsRemaining": ""
```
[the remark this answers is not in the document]
kberg-amzn: Corrected

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -5,120 +5,142 @@
...
-    }
+        const AZ::Vector3& position = GetEntity()->GetTransform()->GetWorldTM().GetTranslation();
+        const HitEffect& effect = GetHitEffect();
 
-    void EnergyBallComponentController::OnCollisionBegin(const AzPhysics::CollisionEvent& collisionEvent)
-    {
-        if (collisionEvent.m_body2)
+        IntersectResults results;
+        const ActivateEvent activateEvent{ GetEntity()->GetTransform()->GetWorldTM(), position, Multiplayer::InvalidNetEntityId, Multiplayer::InvalidNetEntityId };
+        GatherEntities(GetGatherParams(), activateEvent, m_filteredNetEntityIds, results);
```
AMZN-alexpete, line 94: it's probably unlikely, but if this is not doing a shape cast from the old to new position, it's possible a player might run at an energy ball and it not hit depending on the player speed and energy ball speed.  so long as the ball and player move slow enough we should be good.

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -5,120 +5,142 @@
...
+                    float hitDistance = 1.f;
+                    float maxDistance = 1.f;
+                    float damage = effect.m_hitMagnitude * powf((effect.m_hitFalloff * (1.0f - hitDistance / maxDistance)), effect.m_hitExponent);
+
+                    // Look for physics rigid body component and make impact updates
+                    if (Multiplayer::NetworkRigidBodyComponent* rigidBodyComponent = handle.GetEntity()->FindComponent<Multiplayer::NetworkRigidBodyComponent>())
+                    {
+                        const AZ::Vector3 explosionCentre = position;
+                        const AZ::Vector3 hitObject = handle.GetEntity()->GetTransform()->GetWorldTM().GetTranslation();
+                        const AZ::Vector3 impulse = (hitObject - position).GetNormalized() * damage * sv_EnergyBallImpulseScalar;
```
AMZN-alexpete, line 115: when a player runs at an energy ball headed straight for them, if the intersection tunnels through the player, the impulse might accelerate them forward

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
kberg-amzn: Yeah, tunneling could definitely be a potential issue, but the ball does move quite slowly typically. I've thrown a video of what I've configured into the PR description, including the revised jump-pad effect, but design can tune this even lower if they want players to try more skill-based jumping and dodging of these traps.

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
kberg-amzn: Right now this doesn't affect players, only dynamic rigids in the world (like the boxes in the video). Players only receive damage. We could try adding velocity to the players as well if we wanted a quick and dirty simulation of knock-back, but I decided not to press my luck with this PR..

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -5,120 +5,142 @@
...
+            bool shouldTerminate = false;
+            for (const IntersectResult& result : results)
             {
-#if AZ_TRAIT_SERVER
-                if (NetworkHealthComponent* health = target->FindComponent<NetworkHealthComponent>())
+                Multiplayer::ConstNetworkEntityHandle handle = Multiplayer::GetMultiplayer()->GetNetworkEntityManager()->GetEntity(result.m_netEntityId);
+                if (handle.Exists())
                 {
-                    health->SendHealthDelta(aznumeric_cast<float>(-m_armorDamage));
+                    shouldTerminate = true;
```
AMZN-alexpete, line 103: does this mean an energy ball will not terminate if it hits some world geometry or terrain that doesn't have a network entity handle?

AMZN-alexpete: do we have some kind of range for these to prevent them from living/ticking forever?

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
mbalfour-amzn: i was more concerned about the opposite, which is that it will tick twice in the same frame every couple frames in the default case, which seems pretty wasteful, and we've got other cvars exposed that could accidentally cause this to be aligned either better _or_ worse. feels like at a minimum this should be exposed as a cvar too, but i'm not super concerned about this since it's just an energy ball, not a reusable system.

Gem/Code/Source/AutoGen/NetworkPlayerMovementComponent.AutoComponent.xml
```diff
@@ -33,6 +33,30 @@
...
+        Name="SetVelocity"
+        InvokeFrom="Server"
+        HandleOn="Authority"
+        IsPublic="false"
+        IsReliable="true"
+        GenerateEventBindings="true"
+        Description="Directly sets the self generated velocity to the provided value, overriding any current velocity values">
+        <Param Type="AZ::Vector3" Name="Velocity"/>
+        <Param Type="bool" Name="External" Description="If true, the impulse will be applied to the external velocity, and if false it will be applied to internal velocity" />
+    </RemoteProcedure>
```
mbalfour-amzn, line 58: Nit: The bool is probably unnecessary, since anything wanting to set the player velocity through an RPC is "external" by definition i would think, so it always ought to be true. 

Not harmful to have it here though, other than more network data and needing to look up what "true" means when looking at the calling code.

mbalfour-amzn (approved):

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
kberg-amzn: Yup, that's correct.. it bounces because of the rigid component, which might allow for some interesting gameplay behaviours.

AMZN-Olex (approved):

PR #268, opened 2023-03-27
Reticle flashes only on player who shot (No C++ changes)

AMZN-Olex:
- Inserted a node to check if the local player is autonomous and only then play the reticle effect

image

- Only minor cleanup for player damaged effect canvas


kberg-amzn (approved):
Nice

PR #275, opened 2023-03-28
Sprint toggle

kberg-amzn:
Makes the shift key toggle player sprint state, rather than require the shift key to be held down to sprint.

### DIFF
--- a/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
+++ b/Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
@@ -137,12 +137,6 @@ namespace MultiplayerSample
         // Inputs for your own component always exist
         NetworkPlayerMovementComponentNetworkInput* playerInput = input.FindComponentInput<NetworkPlayerMovementComponentNetworkInput>();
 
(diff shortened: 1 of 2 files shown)


Gem/Code/Source/Components/NetworkPlayerMovementComponent.cpp
```diff
@@ -137,6 +137,12 @@ namespace MultiplayerSample
         // Inputs for your own component always exist
         NetworkPlayerMovementComponentNetworkInput* playerInput = input.FindComponentInput<NetworkPlayerMovementComponentNetworkInput>();
 
+        if (m_toggleSprint)
+        {
+            m_sprinting = !m_sprinting;
+            m_toggleSprint = false;
+        }
+
```
mbalfour-amzn, line 145: Should the toggle move into the "if PlayerActionsAllowed() == All" block below (line 160ish)? It seems a bit weird to toggle when the player can't move...

kberg-amzn: oh.. yup I can make that change 👍

galibzon (approved):

PR #277, opened 2023-03-29
Added several energy ball traps, updated gem spawn locations 

mcphedar:
to fit with traps, redid the lower round area in anticipation of using jump pads to let players jump up, moved teleport exits out of dangerous areas


AMZN-Gene (approved):
Fix PR title and description

mbalfour-amzn (approved):

PR #285, opened 2023-03-29
Adding First Match Delay CVar

AMZN-Gene:
Adding _sv_MpsFirstMatchDelaySeconds_ to allow devs to delay when a tournament starts.
_sv_MpsFirstMatchDelaySeconds_ is the delay time AFTER the first player has connected.

Note: This doesn't support changing sv_MpsFirstMatchDelaySeconds cvar at runtime after the first player has already joined.

image


Gem/Code/Source/GameState/GameStateWaitingForPlayers.cpp
```diff
@@ -8,9 +8,16 @@
 #include <GameState/GameStateRequestBus.h>
 #include <Source/GameState/GameStatePreparingMatch.h>
 #include <Source/GameState/GameStateWaitingForPlayers.h>
+#include <PlayerMatchLifecycleBus.h>
 
 namespace MultiplayerSample
 {
+    AZ_CVAR(uint32_t, sv_MpsFirstMatchDelaySeconds, 0, nullptr, AZ::ConsoleFunctorFlags::DontReplicate,
+        "Controls how many seconds the server waits to start the first match after the first player has connected. "
+        "This is a cvar instead of network archetype because it's important for server admins to be able to change this value with commandline instead of having to update a level prefab. "
```
lmbr-pip, line 17:
```suggestion
        "This is a cvar instead of network archetype because it's important for server admins to be able to set this as a launch parameter instead of having to update a level prefab. This setting could be removed if using holding lobby and matchmaking."
```

lmbr-pip (approved):

kberg-amzn (approved):

PR #295, opened 2023-03-30
Fixed dependencies XML file to work with the correct path based wildcard resolving logic.

AMZN-stankowi:
This fixes the MultiplayerSample_Dependencies.xml to properly find all dependencies with the now-corrected wildcard path handling, as seen here: https://github.com/o3de/o3de/pull/15486

This works with and without that other pull request.


mbalfour-amzn (approved):
I kinda miss being able to do recursive subdirectory includes, i'm a little worried about keeping this list maintained. But thanks for finding and fixing the issue here!

MultiplayerSample_Dependencies.xml
```diff
@@ -1,6 +1,22 @@
 <EngineDependencies versionnumber="1.0.0">
     <Dependency path="*.azshadervariant*" optional="false" /> <!-- The wildcard includes azshadervarianttree assets too -->
```
mbalfour-amzn, line 2: Actually, isn't this line broken too? The azshadervariant files exist in multiple subdirectories.

mbalfour-amzn (changes requested):

MultiplayerSample_Dependencies.xml
AMZN-stankowi: I'll double check when this build finishes, but I thought that was matching differently because there were no directory markers, so it was just matching on file name.

AMZN-stankowi:
> I kinda miss being able to do recursive subdirectory includes, i'm a little worried about keeping this list maintained. But thanks for finding and fixing the issue here!

I went back and forth on this. One of the other options was to add support to the path matching function to include the double wildcard "**" for matching multiple folders.

I decided on this solution for these reasons:
1. It's simpler, so it's a less risky change into stabilization.
2. It doesn't prevent us from adding that later, if path is upgraded to support **, then we can change the dependency file again.
3. The bigger issue here is that there are files not emitting product dependencies - the best solution here is to make sure all assets properly emit product dependencies, so we don't have to wildcard pull in a bunch of files that may not be in actual use.

lmbr-pip (approved):

MultiplayerSample_Dependencies.xml
AMZN-stankowi: It looks like it's picking them up. I'm seeing a lot of shader variants:

image

The scan folder path to that first one is:
Shaders/DiffuseGlobalIllumination/diffuseprobegridvisualizationraytracingmiss_dx12_0.azshadervariant

mbalfour-amzn (approved):

lmbr-pip:
This work should go to stabilization as thats where all development is currently based.

PR #299, opened 2023-03-31
Add gem spawning on death

mbalfour-amzn:
When a player dies, they will now spawn a gem cluster where they died worth the same amount that they lose upon death.

Design notes:
* A gem will be spawned from a player's health going to 0, no matter the cause (other player shooting, energy trap, etc).
* The gem will be spawned at the player's position.
* The gem will be worth the exact amount that the dying player loses, or the value set for the gem in the gem table, whichever is higher. (Current configuration sets the gem cluster to 0)
* A gem that is worth 0 will not spawn.
* The gem will despawn at the end of the round if nobody has picked it up.
* The gem will be configured by listing a gem tag on the Network Match component for which gem type to spawn
* The gem tag will be expected to match an entry in the Gem Spawner component in the "Gem Spawnables" table to determine the gem prefab to spawn and the minimum value of the spawned gem.

This PR also has a couple of small cleanups:
* Fixes the physics bus used to teleport a player. The previous bus caused the player to exist for an extra frame by changing the physics representation incorrectly, so the dying player would immediately pick up the spawned gem.
* Small bit of polish to despawn all the gems at round end, as opposed to doing a despawn/spawn both at round start. it just looks a bit nicer IMO.

image
image

### DIFF
--- a/Gem/Code/Source/AutoGen/GemSpawnerComponent.AutoComponent.xml
+++ b/Gem/Code/Source/AutoGen/GemSpawnerComponent.AutoComponent.xml
@@ -26,14 +26,14 @@
                        ExposeToEditor="true" Description="A tag value for entities the describe the location of gems spawn points." />
 
     <RemoteProcedure Name="RPC_SpawnGem" InvokeFrom="Server" HandleOn="Authority" 
-                     IsPublic="true" IsReliable="true" GenerateEventBindings="false" Description="Spawn a gem at a specific location">
+                     IsPublic="true" IsReliable="true" GenerateEventBindings="true" Description="Spawn a gem at a specific location">
       <Param Type="Multiplayer::NetEntityId" Name="PlayerEntity"/>
       <Param Type="AZ::Vector3" Name="SpawnLocation"/>
       <Param Type="AZStd::string" Name="GemTag"/>
     </RemoteProcedure>
 
   <RemoteProcedure Name="RPC_SpawnGemWithValue" InvokeFrom="Server" HandleOn="Authority"
-                   IsPublic="true" IsReliable="true" GenerateEventBindings="false" Description="Spawn a gem at a specific location with the given value">
+                   IsPublic="true" IsReliable="true" GenerateEventBindings="true" Description="Spawn a gem at a specific location with the given value">
     <Param Type="Multiplayer::NetEntityId" Name="PlayerEntity"/>
     <Param Type="AZ::Vector3" Name="SpawnLocation"/>
     <Param Type="AZStd::string" Name="GemTag"/>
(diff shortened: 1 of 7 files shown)


AMZN-Gene (approved):

Gem/Code/Source/AutoGen/GemSpawnerComponent.AutoComponent.xml
```diff
@@ -24,4 +24,20 @@
 
     <ArchetypeProperty Type="AZStd::string" Name="GemSpawnTag"
                        ExposeToEditor="true" Description="A tag value for entities the describe the location of gems spawn points." />
+
+    <RemoteProcedure Name="RPC_SpawnGem" InvokeFrom="Server" HandleOn="Authority" 
+                     IsPublic="true" IsReliable="true" GenerateEventBindings="false" Description="Spawn a gem at a specific location">
```
kberg-amzn, line 29: Maybe set the generate event bindings to true so users can spawn gems from script too?

Levels/NewStarbase/NewStarbase.prefab
```diff
@@ -5538,7 +5538,8 @@
                     "Id": 110022581788315572,
                     "m_template": {
                         "$type": "MultiplayerSample::NetworkMatchComponent",
-                        "RespawnPenaltyPercent": 20
+                        "RespawnPenaltyPercent": 20,
```
kberg-amzn, line 5541: Make sure this merges with Darrin's level revisions.. I think he had a verbal lock on this file.

mbalfour-amzn: I checked with him, he said i was free to add this

kberg-amzn (approved):
So awesome 👍

Gem/Code/Source/Components/Multiplayer/GemSpawnerComponent.cpp
```diff
@@ -162,28 +162,59 @@ namespace MultiplayerSample
...
+        SpawnGem(spawnLocation, AZ::Crc32(gemTag));
+    }
+
+    void GemSpawnerComponentController::HandleRPC_SpawnGemWithValue(
+        [[maybe_unused]] AzNetworking::IConnection* invokingConnection,
+        [[maybe_unused]] const Multiplayer::NetEntityId& playerEntity, 
+        const AZ::Vector3& spawnLocation, const AZStd::string& gemTag, const uint16_t& gemValue)
+    {
+        auto gemEntry = GetGemSpawnable(AZ::Crc32(gemTag));
+        if (gemEntry)
```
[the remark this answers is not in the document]
mbalfour-amzn: it's specifically an optional<GemSpawnable>, which was a bit cumbersome and why i chose auto.

PR #300, opened 2023-03-31
Add Time Delay Between Matches

AMZN-Gene:
- Adding network variable to track time delay between matches
- Increased time delay between rounds
- Update match over HUD to display the time until a new match starts
- Moved GameState code to only compile on server; 1) Help alleviate confusing thinking that game states run on both client and server and 2) removes the need to use AZ_TRAIT_SERVER within game state code when using some component controller functions.

image


Gem/Code/Source/Components/UI/UiGameOverComponent.cpp
```diff
@@ -59,23 +59,33 @@ namespace MultiplayerSample
         m_onRoundNumberChangedHandler.Disconnect();
     }
 
-    void UiGameOverComponent::DisplaySecondsRemainingUI(uint16_t secondsRemaining)
+    void UiGameOverComponent::DisplaySecondsRemainingUI()
     {
+        const AZ::TimeMs currentHostTime = AZ::Interface<Multiplayer::IMultiplayer>::Get()->GetCurrentHostTimeMs();
```
kberg-amzn, line 64: Nice 👍

kberg-amzn (approved):
Interesting changes.. nice that you were able to leverage the synced host time value and also remove all the game state logic from the client because both of those are good examples for how to structure these features in a multiplayer game. The malfunctioning shield generator might interact weirdly? possibly? because it's watching the current round time and total round time and storing off an internal lifetime value to track when and how often to go boom, but it should be an easy fix if so, and I want to to submit and get these changes out first before digging in.

mbalfour-amzn (approved):

PR #317, opened 2023-04-05
PopcornFX Assets README

micronAMZN:
Added README for PopcornFX assets based on the presentation supplied by PopcornFX. Added media (png and gif) for the README. Edited assetprocessor_settings.setreg so that the media folder is ignored by AP.


PopcornFX/README.md
```diff
@@ -0,0 +1,207 @@
+# PopcornFX Assets - O3DE Multiplayer Sample
+
+The Multiplayer Sample includes nine visual FX assets created with PopcornFX. This README documents the these FX assets so that you can customize them or deconstruct them to learn how to build PopcornFX assets of your own.
```
lmbr-pip, line 3: Is nine relevant? I would avoid specifying actual number so we don't have to keep updating the document.

FX, while common term in game industry, should be defined on first use.

May also want to call out not all of these FX are currently used in MPS.

PopcornFX/README.md
```diff
@@ -0,0 +1,207 @@
...
+* **LyShine**
+* **Starting Point Input**
+
+## General information
+
+Some of the FX are split into multiple parts (for example, Blast, Projectile, and Explosion). Splitting the FX makes it easier to integrate the FX into varied gameplay scenarios.
+
+Many of the FX have a **Color Intensity** property that you can modify to adjust the emissive effect. The emissive effect requires that the **PostFX Layer** and **Bloom** components are added to the scene.
+
+<img src="readme_media/image1.png">
```
lmbr-pip, line 25: Images don't work in preview. Should we be using markdown versions ie alt text?

micronAMZN: The problem is LFS. I'm using img tags because I need to size some images inside of tables. I've modified the `.gitattributes`, but need to see what I can do about drastically reducing media sizes.

PopcornFX/README.md
micronAMZN: Changed nine to "several"

Changed to consistently use the acronym "VFX" and define it after first use.

I'm hesitant to call out assets as not being used because there are feature requests to implement them.

PopcornFX/README.md
micronAMZN: This should all be fixed. I have to use GIFs, but I have crunched them down to the smallest possible size while maintaining decent quality. Because I resized all the images I was able to switch to standard markdown tags for images.

mcphedar (approved):

kberg-amzn (approved):

Issue #313, opened 2023-04-05
MultiplayerSample Project README.md provides incorrect controls


LB-MichalRodzos:
**Describe the bug**
In section "Player Controls" of the MultiplayerSample Project README.md is written:

Sprint: Hold `shift`

However; Shift key works as a toggle between Walk and Run.

Please note that the issue should be considered a Minor.

**Steps to reproduce**
1. Go to https://github.com/o3de/o3de-multiplayersample/#multiplayersample-project.
2. Follow the instruction to download and install the MultiplayerSample project.
3. Run the NewStarbase level (in Editor's Game Mode or GameLauncher).
4. While moving the character, press the Shift button.
5. Go back to the MultiplayerSample Project README.md.
6. Find Player Controls.

**Expected behavior**
Player Controls in the MultiplayerSample Project README.md shows correct controls.

**Actual behavior**
Player Controls in the MultiplayerSample Project README.md shows incorrect controls for running.

**Found in Branch**
Stabilization/2305 Installer build (https://o3debinaries.org/stabilization-2305/Latest/Windows/o3de_installer.exe, https://o3debinaries.org/stabilization-2305/Latest/Linux/O3DE_latest.deb)

**Desktop/Device:**

- Device: PC
- OS: Windows
- Version: 10 Pro
- CPU: AMD Ryzen 5 3600 6-Core Processor
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Memory: 16 GB

lmbr-pip:
Good catch, now we need to just toggle `shift` to switch sprinting mode. Will update.

lmbr-pip:
Fixed in https://github.com/o3de/o3de-multiplayersample/blob/stabilization/2305/README.md

LB-MichalRodzos:
Issue verified as fixed in https://github.com/o3de/o3de-multiplayersample/blob/stabilization/2305/README.md. @lmbr-pip Will this fix also be merged to development documentation (https://github.com/o3de/o3de-multiplayersample/#multiplayersample-project) as the issue still occurs there?

PR #325, opened 2023-04-06
Fix documentation of sprint control

lmbr-pip:
Addresses https://github.com/o3de/o3de-multiplayersample/issues/313


lmbr-pip:
Incorporated into latest changes in https://github.com/o3de/o3de-multiplayersample/blob/stabilization/2305/README.md

PR #345, opened 2023-04-12
Adding first-pass code for a partially functioning settings screen.

mbalfour-amzn:
With this code, here's what works and what doesn't:
* Graphics API toggle will default to the current API, but changing it does nothing.
* Texture Quality will default to the current quality and can be changed, but the changes don't persist between runs.
* Master Volume will default to 100% and can be changed, but the changes don't persist between runs.

image


Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -0,0 +1,230 @@
...
+    {
+        const char* labels[] =
+        {
+            "Ultra (4K)",
+            "High (2K)",
+            "Medium (1K)",
+            "Low (512)",
+            "Very Low (256)",
+            "Extremely Low (128)",
+            "Rock Bottom (64)"
```
AMZN-Olex, line 177: Nice

AMZN-Olex (approved):

AMZN-alexpete (approved):

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -0,0 +1,230 @@
...
+{
+    void MpsSettings::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<MpsSettings>()
+                ->Version(0)
+                ->Field("GraphicsApi", &MpsSettings::m_atomApiType)
+                ->Field("MasterVolume", &MpsSettings::m_masterVolume)
+                ->Field("TextureQuality", &MpsSettings::m_streamingImageMipBias)
```
AMZN-alexpete, line 28: can we add a setting for turning off "Pain" sounds? hehe, or maybe that's not a thing anymore :)

mbalfour-amzn: Oh, no, I can't add that. That setting wouldn't make sense for most people, we only play the pain sound on your pc...
```
if (getenv("LOGNAME") == "AMZN-alexpete")
{
    TriggerSound("PAIN!!!!");
}
```

PR #354, opened 2023-04-12
Remove the entry player spawner entry causing players to spawn into space

mbalfour-amzn:
If the player spawner system finds an empty entry, it will spawn the player using an uninitialized transform. A separate PR will be put up to make that code more robust.


kberg-amzn (approved):
We also probably need a GHI to go back and refactor the simple player spawner soonish.

AMZN-Olex (approved):

PR #363, opened 2023-04-17
Energy Ball Fixes

### LINKED ISSUE #359: Space Cannon Energy Ball projectiles are not always visible
**Describe the bug**
In the 4/10 playtest. The space cannon's energy ball projectiles cannot always be seen. This issue is inconsistent between players, with some seeing the balls as they fly and others being unable to. Their visibility also depends on the world position and potentially the direction they are fired in. All players can see the balls forming at the energy cannon, but when the projectile VFX starts, they disappear. Sometimes, upon player respawn or reload the balls became visible again.

**Steps to reproduce**
1. Build o3de-multiplayersample
2. Open the NewStarBase level in the editor
3. Play the game (ctrl-g)
4. Observe the various energy cannons placed around the level. Some may fire visible projectiles, but most will not.

**Expected behavior**
You should be able to see the projectiles as they travel

**Actual behavior**
Visibility is inconsistent

**Screenshots/Video**
(https://user-images.githubusercontent.com/67011188/232088761-6d97b8d3-eb13-404f-9a2a-5437aad63972.mp4)

**Found in Branch**
03de-multiplayersample stabilization/2305

mbalfour-amzn:
This PR fixes #359 .

There are a number of different fixes contained in this PR:
* When transitioning in and out of game mode in the Editor, order of operation issues can cause PopcornFX to delete the m_emitter pointer before GameEffect is done trying to use it. The IsEffectAlive() checks added to GameEffect detect when this case occurs and prevents the pointer from getting dereferenced and crashing. The asserts are there to continue to notify about the underlying problem, which is that we need to find a more foolproof way to know that PopcornFX has deleted the m_emitter pointer.
* The root cause of the disappearing energy balls was that if Kill() got called on a particle effect in the first frame, the particle effect would start to delete itself, but never complete the deletion, so subsequent Restart() calls would never revive it. Kill() could sometimes get called in the first frame depending on RPC timing. This replaces Kill() with Terminate() as a workaround to prevent that from happening.
* The energy ball should call Restart() instead of Start(). Start() will fail if the the particle effect is still finishing its on death animation. Restart will wait for it to finish then restart it.
* The BallExplosion / LaunchBall events weren't guaranteed to arrive in order, so the logic wasn't always necessarily managing the particle state correctly. By switching to a "BallActive" property and removing "LaunchBall", the order of operations problem goes away.
* The physics call to disable/enable physics should use the SimulatedBody bus, not the RigidBody bus.
* There were multiple redundant calls to BallExplosion and the collision checking that could happen during a single launched ball due to bad state management. This adds checks for "is ball active" server-side to avoid the redundant calls.
* Added @kberg-amzn 's optional debug draws to make it easier to visualize network problems vs particle problems and touched it up a bit so that it draws a single sphere that's the same size as the physics representation instead of a hard-coded size and multiple spheres.


AMZN-Olex (approved):

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -37,39 +40,121 @@ namespace MultiplayerSample
...
     void EnergyBallComponent::OnDeactivate([[maybe_unused]] Multiplayer::EntityIsMigrating entityIsMigrating)
     {
+#if AZ_TRAIT_CLIENT
+        AZ::TickBus::Handler::BusDisconnect();
+#endif
     }
 
-#if AZ_TRAIT_CLIENT
-    void EnergyBallComponent::HandleRPC_BallLaunched([[maybe_unused]] AzNetworking::IConnection* invokingConnection, [[maybe_unused]] const AZ::Vector3& location)
+    void EnergyBallComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
```
kberg-amzn, line 58: You could actually, if you wanted, attach an event handler to the BallActive network property so that when the value changes your callback gets invoked. Since it's just an on/off thing I think it's safe to just say `if (active) startfx else killfx` although you've clearly run into some hair pulling bugs with pkfx.

Only benefit is if somebody goes crazy and adds 100,000 energy balls to a level the game doesn't have to go through and tick all 100,000 every frame on the client and the server, you'd only process effects changes when the state of BallActive changes.

mbalfour-amzn: ooh, neat, i didn't know the "on changed" events existed! i'll switch over to using that.

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.h
```diff
@@ -25,17 +23,26 @@ namespace MultiplayerSample
...
+
+        AZ::ScheduledEvent m_debugDrawEvent{ [this]()
+        {
+            DebugDraw();
+        }, AZ::Name("EnergyBallDebugDraw") };
+
+        AZ::Event<bool>::Handler m_ballActiveHandler{ [this](bool active)
+        {
+            OnBallActiveChanged(active);
+        } };
```
AMZN-alexpete, line 43: out of curiosity is there some format that lets you forgo the lambda and do something like
```
AZ::Event<bool>::Handler m_ballActiveHandler = &EnergyBallComponent::OnBallActiveChanged;
```
?

mbalfour-amzn: Sort of? 

I think I'd have to make it a static method instead of a class method and make it figure out which instance is affected to work around the lack of a "this" pointer, so the lambda seems like a tidier solution unless you know a trick I'm overlooking.

PR #368, opened 2023-04-19
Adds background music to the multiplayer sample project

kberg-amzn:
Adds three new streaming audio pieces composed by the legendary Kevin MacCloud:
- Rocket
- Future Gladiator
- Beauty Flow

Adds a new background music component to the multiplayer sample, and adds that component to the AmbientAudio entity on the level:
image

The component will loop through the user-defined playlist playing each track after the next in an infinite loop.

### DIFF
--- a/Gem/Code/Source/Components/BackgroundMusicComponent.cpp
+++ b/Gem/Code/Source/Components/BackgroundMusicComponent.cpp
@@ -20,6 +20,7 @@ namespace MultiplayerSample
         if (serializeContext)
         {
             serializeContext->Class<BackgroundMusicComponent, AZ::Component>()
+                ->Field("Shuffle", &BackgroundMusicComponent::m_shuffle)
                 ->Field("Playlist", &BackgroundMusicComponent::m_playlist)
                 ->Version(1);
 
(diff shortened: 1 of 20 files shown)


mcphedar (approved):

Gem/Code/Source/Components/BackgroundMusicComponent.cpp
```diff
@@ -0,0 +1,98 @@
...
+#if AZ_TRAIT_CLIENT
+        m_audioSystem = AZ::Interface<Audio::IAudioSystem>::Get();
+
+        if (m_audioSystem != nullptr)
+        {
+            m_trackIndex = size_t(-1);
+            ReportTriggerFinished(INVALID_AUDIO_CONTROL_ID);
+        }
+#endif
+
```
mbalfour-amzn, line 64: Suggestion: Consider initializing m_trackIndex to a random number % playlist.size() on each Activate(). Could potentially even make this an optional checkbox on the component. But without it, anyone who tests this game a bunch will hear the first song far more often than the other. ;) 

A slightly bigger variation would be a "shuffle mode" that plays every song in random order exactly once before repeating. But I'd personally be happy with at least "don't always start at the start".

https://github.com/RondeSC/Miller_Shuffle_Algo

mbalfour-amzn (approved):

Gem/Code/Source/Components/BackgroundMusicComponent.cpp
kberg-amzn: Yeah, that's a super good call.. that's the exact sort of thing that will drive developers completely nuts.

PR #374, opened 2023-04-20
Add persistent user settings support.

mbalfour-amzn:
Thanks to some significant help from @lumberyard-employee-dm , this PR adds load/save support to MPS user settings through use of a loose .setreg file. This has been tested with loose file builds as well as pak builds.

Implementation was a little bit tricky due to the user setting of "graphics API", which requires the user settings to get loaded and applied before any system components started up. Thanks to @lumberyard-employee-dm for finding the happy location of the MPS Module constructor to make this work! Thanks also to @lumberyard-employee-dm for providing code to show how to make user settings work with a .setreg file in a secure way that ensures that no engine settings can be directly modified by loading it into its own registry subkey.

One improvement we can make to this code in development instead of stabilization is to add an API to Atom's Factory Manager to enumerate all of the existing factories, so that the user settings toggle will only show available factories instead of all factories. It's unfortunately too late to get an engine change like that into stabilization though, so for now the toggle will show all choices regardless of what's available, but only available choices will work. Selecting an unavailable choice (like "metal" on a PC) will just cause Atom to select its default API choice (dx12 in this example).

### DIFF
--- a/Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
+++ b/Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
@@ -67,7 +67,7 @@ namespace MultiplayerSample
 
     void UiSettingsComponent::Activate()
     {
-        // Make sure to load and apply the current user settings when this component activates.
+        // Loads and applies the current user settings when this component activates.
         // The user settings should *already* be loaded and applied at Launcher startup, but connecting to the server
         // and switching levels can cause some engine settings to reset themselves, so this will reapply the desired
         // user settings again.
--- a/Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
+++ b/Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
@@ -6,30 +6,16 @@
  */
 
 #include <Atom/RHI/Factory.h>
-#include <AzCore/Console/IConsole.h>
 #include <AzCore/Interface/Interface.h>
 #include <AzCore/Serialization/EditContext.h>
-#include <IAudioSystem.h>
 #include <LyShine/Bus/UiButtonBus.h>
 #include <LyShine/Bus/UiElementBus.h>
 #include <LyShine/Bus/UiTextBus.h>
 #include <Source/Components/UI/UiSettingsComponent.h>
+#include <Source/UserSettings/MultiplayerSampleUserSettings.h>
 
 namespace MultiplayerSample
 {
-    void MpsSettings::Reflect(AZ::ReflectContext* context)
-    {
-        if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
-        {
-            serializeContext->Class<MpsSettings>()
-                ->Version(0)
-                ->Field("GraphicsApi", &MpsSettings::m_atomApiType)
-                ->Field("MasterVolume", &MpsSettings::m_masterVolume)
-                ->Field("TextureQuality", &MpsSettings::m_streamingImageMipBias)
-                ;
-        }
-    }
-
     void UiToggle::Reflect(AZ::ReflectContext* context)
     {
         if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
@@ -40,7 +26,6 @@ namespace MultiplayerSample
                 ->Field("LeftButton", &UiToggle::m_leftButtonEntity)
                 ->Field("RightButton", &UiToggle::m_rightButtonEntity)
                 ;
-
 
             if (AZ::EditContext* editContext = serializeContext->GetEditContext())
             {
@@ -56,7 +41,6 @@ namespace MultiplayerSample
 
     void UiSettingsComponent::Reflect(AZ::ReflectContext* context)
     {
-        MpsSettings::Reflect(context);
         UiToggle::Reflect(context);
 
         if (AZ::SerializeContext* serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
@@ -83,20 +67,11 @@ namespace MultiplayerSample
 
     void UiSettingsComponent::Activate()
     {
-        // Initialize our user settings 
-
-        // Initialize the current streaming image mip bias setting.
-        if (AZ::IConsole* console = AZ::Interface<AZ::IConsole>::Get(); console)
-        {
-            int16_t mipBias = 0;
-            console->GetCvarValue("r_streamingImageMipBias", mipBias);
-            m_settings.m_streamingImageMipBias = aznumeric_cast<uint8_t>(mipBias);
-        }
-
-        // Initialize the graphics API type
-        m_settings.m_atomApiType = AZ::RHI::Factory::Get().GetAPIUniqueIndex();
-
-        // There's currently no way to initialize the master volume, this doesn't seem to be fetchable anywhere.
+        // Make sure to load and apply the current user settings when this component activates.
+        // The user settings should *already* be loaded and applied at Launcher startup, but connecting to the server
+        // and switching levels can cause some engine settings to reset themselves, so this will reapply the desired
+        // user settings again.
+        MultiplayerSampleUserSettingsRequestBus::Broadcast(&MultiplayerSampleUserSettingsRequestBus::Events::Load);
 
         // Initialize the toggles to the current values
         OnGraphicsApiToggle(ToggleDirection::None);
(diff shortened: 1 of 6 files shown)


Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -83,20 +67,11 @@ namespace MultiplayerSample
...
-            int16_t mipBias = 0;
-            console->GetCvarValue("r_streamingImageMipBias", mipBias);
-            m_settings.m_streamingImageMipBias = aznumeric_cast<uint8_t>(mipBias);
-        }
-
-        // Initialize the graphics API type
-        m_settings.m_atomApiType = AZ::RHI::Factory::Get().GetAPIUniqueIndex();
-
-        // There's currently no way to initialize the master volume, this doesn't seem to be fetchable anywhere.
+        // Make sure to load and apply the current user settings when this component activates.
```
lmbr-pip, line 70: Comment seems to say developer needs to take action, but then says component will reapply the desired settings.

So is this more "Loads and applies the current user setting when this component..."?

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -140,91 +115,132 @@ namespace MultiplayerSample
...
+    uint32_t UiSettingsComponent::GetRotatedIndex(
+        const AZStd::vector<AZStd::pair<ValueType, AZStd::string>>& valuesToLabels,
+        const ValueType& value, ToggleDirection toggleDirection)
+    {
+        const size_t totalValues = valuesToLabels.size();
+
+        uint32_t curIndex = 0;
+
+        // Loop through and look for the correct value
+        for (size_t index = 0; index < totalValues; index++)
```
lmbr-pip, line 128: nit: use a range loop instead ?

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+        m_userSettingsPath /= "Registry";
+        m_userSettingsPath /= "MultiplayerSampleUserSettings.setreg";
+
+        // Load all of our settings keys, create default values if they don't exist and initialize the engine settings as appropriate.
+        Load();
+    }
+
+	MultiplayerSampleUserSettings::~MultiplayerSampleUserSettings()
+	{
+        MultiplayerSampleUserSettingsRequestBus::Handler::BusDisconnect();
```
lmbr-pip, line 42: nit: whitespace looks off here

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+            if (audioSystem)
+            {
+                Audio::TAudioObjectID rtpcId = audioSystem->GetAudioRtpcID("Volume_Master");
+
+                if (rtpcId != INVALID_AUDIO_CONTROL_ID)
+                {
+                    Audio::ObjectRequest::SetParameterValue setParameter;
+                    setParameter.m_audioObjectId = INVALID_AUDIO_OBJECT_ID;
+                    setParameter.m_parameterId = rtpcId;
+                    setParameter.m_value = masterVolume / 100.0f;
```
lmbr-pip, line 139: nit: add comment that this can be between 0 and 100, where 0 is off and 100 is full volume?

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.h
```diff
@@ -0,0 +1,90 @@
...
+    // This implements the bus provided above. The user settings get auto-loaded at construction and auto-saved at destruction,
+    // though saves can also be triggered at other times as well. Because one of the settings is the default graphics API, these
+    // settings need to be loaded before system components are initialized because the Atom system components load the graphics
+    // API. All of the other settings are changeable at any time and would have allowed this class to get created later in the
+    // boot process.
+	class MultiplayerSampleUserSettings : public MultiplayerSampleUserSettingsRequestBus::Handler
+	{
+	public:
+		MultiplayerSampleUserSettings();
+		~MultiplayerSampleUserSettings();
```
lmbr-pip, line 60: override on destructor?

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -140,91 +115,132 @@ namespace MultiplayerSample
...
+        default:
+            return curIndex;
+        }
+    }
+
     void UiSettingsComponent::OnGraphicsApiToggle(ToggleDirection toggleDirection)
     {
         // This list is expected to match the values in AZ::RHI::ApiIndex.
-        const char* labels[] =
+        const AZStd::vector<AZStd::pair<AZStd::string, AZStd::string>> valuesToLabels =
```
kberg-amzn, line 151: Out of curiosity, no action needed here, but do we have a GHI for Atom to maintain this mapping in the future so that adding a new rendering backend doesn't require modifications to customer product code?

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -140,91 +115,132 @@ namespace MultiplayerSample
...
+        const AZStd::vector<AZStd::pair<int16_t, AZStd::string>> valuesToLabels =
         {
-            "Ultra (4K)",
-            "High (2K)",
-            "Medium (1K)",
-            "Low (512)",
-            "Very Low (256)",
-            "Extremely Low (128)",
-            "Rock Bottom (64)"
+            { aznumeric_cast<int16_t>(6), "Rock Bottom (64)" },
```
kberg-amzn, line 185: 🤣

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    AZ::Interface<Audio::IAudioSystem>::Get()->PushRequest(AZStd::move(setParameter));
```
kberg-amzn, line 140: You're a saint to developer ears everywhere..

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.h
```diff
@@ -0,0 +1,90 @@
...
+
+        void Load() override;
+		void Save() override;
```
kberg-amzn, line 63: indentation strikes again..

kberg-amzn (approved):
LGTM! Other than a couple places your editor failed you on indentation, looks like a fantastic change!

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+                // Use SystemFile::Rename to move the file to the final destination
+                userSettingsStream.Close();
+                bool renameSuccess = AZ::IO::SystemFile::Rename(userSettingsStream.GetFilename(), userSettingsSavePath.c_str(), true);
+                AZ_Error("UserSettings", renameSuccess, 
+                    "Renaming '%s' to '%s' failed.", userSettingsStream.GetFilename(), userSettingsSavePath.c_str());
+            }
+        }
+    }
+
+} // namespace MultiplayerSample
```
lumberyard-employee-dm, line 203: protip: IF you delete the last curly brace in this file and then re-add it in Visual Studio, the IDE will automatically fix the indentation.

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: You mean like `for (auto& value : valuesToLabels)`? It's messy to do that when you need the index value (see line 132), since you then need to track the index as a separate incrementing counter. IMO it makes it less readable that just the good old fashioned index loop.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: Agh tabs snuck in. Fixed.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+            if (userSettings.IsSuccess())
+            {
+                // Merge the user settings file under the base "/O3DE/MultiplayerSample/User/Settings" key.
+                // This will ensure that it cannot overwrite any other engine settings.
+                [[maybe_unused]] auto mergeSuccess = registry->MergeSettings(userSettings.GetValue(),
+                    AZ::SettingsRegistryInterface::Format::JsonMergePatch, BaseRegistryKey);
+
+                AZ_Error("UserSettings", mergeSuccess, "Failed to merge user settings into the O3DE registry.");
+            }
```
lumberyard-employee-dm, line 63: There is a SettingsRegistryMergeUtils function called [MergeSettingsToRegistry_ConfigFile](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.h#L189-L200), that can merge a config file directly to the Settings Registry
```suggestion
            AZ::SettingsRegistryMergeUtils::ConfigParser configParser;
            configParser.m_registryRootPointerPath = BaseRegistryKey;
            [[maybe_unused]] auto mergeSuccess = AZ::SettingsRegistryMergeUtils::MergeSettingsToRegistry_ConfigFile(*registry, m_userSettingsPath.Native(), configParser);
            AZ_Error("UserSettings", mergeSuccess, "Failed to merge user settings into the O3DE registry.");
```

lumberyard-employee-dm: Actually this code looks strange.
Why is a Windows style INI/.cfg file being merged to the Settings Registry?

The MergeSettingsFile function can be used directly.
```suggestion
            // Merge the user settings file under the base "/O3DE/MultiplayerSample/User/Settings" key.
            // This will ensure that it cannot overwrite any other engine settings.
            [[maybe_unused]] auto mergeOutcome = registry->MergeSettingsFile(m_userSettingsPath.Native(),
                AZ::SettingsRegistryInterface::Format::JsonMergePatch, BaseRegistryKey);

            AZ_Error("UserSettings", mergeSuccess, "Failed to merge user settings into the O3DE registry with error %s.", mergeOutcome.GetError().c_str());
```

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: Not exactly. I have a change that I can put up for the development branch to add an API to Atom to expose the set of valid rendering backends. I just didn't want to submit it until I move from working on stabilization to working on development. This code would then be able to use that API to get the proper subset of factory names.

That said, the names we would get would be a subset of "null", "dx12", "vulkan", "metal". I'd personally be OK with using those as the display values, but if we wanted fancier names like "DirectX 12", or to filter out "null", we'd still need to do some pretty equivalent label mapping and hard-coding here.

We could theoretically add another API to the factories to get display names separately from factory names, but even then we'd _still_ potentially want to filter out "null", and potentially use different display names (especially if we ever localized), so not sure this can be fully solved the way you'd like.

I'm pretty sure new rendering backends will get added pretty infrequently though.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <AzCore/Settings/SettingsRegistry.h>
+#include <AzCore/Settings/SettingsRegistryMergeUtils.h>
+#include <AzCore/Utils/Utils.h>
+#include <AzFramework/FileFunc/FileFunc.h>
+#include <IAudioSystem.h>
+#include <UserSettings/MultiplayerSampleUserSettings.h>
+
+namespace MultiplayerSample
+{
+	MultiplayerSampleUserSettings::MultiplayerSampleUserSettings()
```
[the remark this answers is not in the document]
mbalfour-amzn: Tabs weirdly snuck in on me.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: I'm not sure that the ConfigParser would work correctly, because this isn't a config file? I admittedly wasn't keen on using GetCfgFileContents() because of the name, but it didn't seem right to rewrite that code here either. I just need the contents of the .setreg file read into memory instead of letting the Settings Registry try to read it because the Settings Registry is using the default FileIOBase, which attempts to read from a PAK file. I needed code that would always read it from a loose file.

I'll add some comments above the GetCfgFileContents() to explain it further.

lumberyard-employee-dm: The [AZ::IO::SystemFile::Length and AZ::IO::SystemFile::Read](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/SystemFile.h#L116-L119) function should be used here to read the file then.

There is no need to try to parse the contents of the file at all.

Gem/Code/Source/MultiplayerSampleModule.cpp
```diff
@@ -72,6 +73,13 @@ namespace MultiplayerSample
                 azrtti_typeid<MultiplayerSampleSystemComponent>(),
             };
         }
+
+#if AZ_TRAIT_CLIENT
+        // This needs to be created as a part of the MultiplayerSampleModule, not during any sort of System Component activation.
+        // It will affect registry keys that get read by System Components as a part of their activation and we can't guarantee
```
AMZN-Gene, line 79: nice comment

AMZN-Gene (approved):

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: I can change it if you feel strongly (though I think I also need to add a trailing \0 after reading, right? and add a ')' to the Length call), but how is putting that implementation inline here any better than using this?

```
        AZ::Outcome<AZStd::string, AZStd::string> GetCfgFileContents(const AZStd::string& filePath)
        {
            if (!AZ::IO::SystemFile::Exists(filePath.c_str()))
            {
                return AZ::Failure(AZStd::string::format("Cfg file '%s' does not exist.", filePath.c_str()));
            }

            // Load the settings file into a string
            AZ::IO::SystemFile settingsFile;
            if (!settingsFile.Open(filePath.c_str(), AZ::IO::SystemFile::SF_OPEN_READ_ONLY))
            {
                return AZ::Failure(AZStd::string::format("Error reading cfg file '%s'.", filePath.c_str()));
            }

            AZStd::string settingsFileContents(settingsFile.Length(), '\0');
            settingsFile.Read(settingsFileContents.size(), settingsFileContents.data());
            settingsFile.Close();
            return AZ::Success(settingsFileContents);

        }
```

It doesn't seem substantially different to me, just more lines of code for me to add to do things that a function in the engine already provides. The only part that's admittedly questionable with my current approach is that "GetCfgFileContents" is a little bit misleading since it isn't a "cfg" file, it's a "setreg" file. But I tried to clarify that in the comments, and maybe this particular method should get renamed in the engine since it's technically capable of reading _any_ text file into memory.

lumberyard-employee-dm: I thought that the GetCfgFileContents function was running a regular expression over each line to parse for CFG content, but I got it confused with the [GetValueForKeyInCfgFileContents](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/FileFunc/FileFunc.cpp#L143-L162) function.

lumberyard-employee-dm: Side Note: For the https://en.cppreference.com/w/cpp/string/basic_string/resize_and_overwrite function, the null-terminating character is taking into account.
The functor that is specified must the size of the data written to the buffer, which is used to null-terminate by the `resize_and_overwrite_function` as the offset to write the [null-terminator and update the string size](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/string/string.h#L1052-L1060)

lumberyard-employee-dm (approved):

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: ohhhh, neat! I've never used resize_and_overwrite before, I didn't realize it handled the null terminator, that's pretty cool!

PR #379, opened 2023-04-21
Add fullscreen/windowed toggle.

mbalfour-amzn:
Adds a fullscreen/windowed toggle to the settings screen. Since ctrl-enter also toggles between fullscreen and windowed modes, the toggle will listen and change to reflect the current state.

This PR also introduces code changes to support a resolution toggle, but the toggle itself is hidden on the screen and the code to apply the resolution change is commented out. While it technically works, there are some UX issues with how it works (Display Setting scaling, changing it while in fullscreen mode, window size vs rendering size, and preventing resolution choices larger than the screen handles), so it's been disabled for now. I left the disabled code in though, because it's still a useful demonstration of how to do it. The resolution toggle will likely become enabled and functional in the development branch, since it will probably need some AzFramework Window Api changes to support it fully.

image


Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -180,6 +188,93 @@ namespace MultiplayerSample
...
+        if (auto* registry = AZ::SettingsRegistry::Get(); registry != nullptr)
+        {
+            if (AZ::IConsole* console = AZ::Interface<AZ::IConsole>::Get(); console)
+            {
+                // This is roughly what it takes to change the screen resolution, but it gets complicated by Dpi scaling
+                // and whether or not the window is currently fullscreen. The code below is temporarily removed until the
+                // logic can be enhanced to account for these other conditions.
+
+                /*
+                // Change the resolution if the window doesn't exist yet.
```
kberg-amzn, line 253: Dead code?

mbalfour-amzn: Yes, but intentionally left in dead code as an example of how to do it if you don't care about perfection. I can accept it if you want me to remove it, I would spiral through and remove all of the functions and toggle code related to supporting resolution then.

kberg-amzn: Maybe a comment pointing folks in the right direction might be better then a chunk of commented out code in that case..
```
// If the window is not yet created, you can set the r_width and r_height console variables to alter the window dimensions
// If the window is created, use the AzFramework::NativeWindowHandle to retrieve the active window handler and then use the AzFramework::WindowRequestBus to alter the dimensions of the active window
```
And also a quick comment as to why you ultimately ended up going down a different path.

kberg-amzn: Ahh, I missed the 'comment why you went a different route' because I was distracted by the dead code 🤣

mbalfour-amzn: lol. It's not a different route or anything yet, it's just that I can't make it work in a user-friendly way without extending our window APIs in AzFramework. I'll continue making this work in development, I'm just limited in how far I can take it in stabilization.

However, given your concern about the dead code, I'm seeing if I can make the code active but not behave differently than now by default.

kberg-amzn: Oh! I think I understand you better now, resolution setting changes from the settings screen are actually unsupported right now with that code removed. With the registry code it actually requires a program restart.. I didn't clue in on that at all. That removes my concerns about dead code, this is actually essential to the correct functioning of the feature.

kberg-amzn: I'm okay with you leaving this commented out here and continuing to iterate on this in dev.

mbalfour-amzn: Ok, it's now active code. You can technically change the resolution in your usersettings.setreg file, but it's not exposed in the UI.

kberg-amzn (approved):

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -235,12 +239,87 @@ namespace MultiplayerSample
...
+            &MultiplayerSampleUserSettingsRequestBus::Events::SetFullscreen, valuesToLabels[fullscreenIndex].first);
+
+        MultiplayerSampleUserSettingsRequestBus::Broadcast(&MultiplayerSampleUserSettingsRequestBus::Events::Save);
+    }
+
+    void UiSettingsComponent::OnResolutionToggle(UiToggle& toggle, ToggleDirection toggleDirection)
+    {
+        const AZStd::vector<AZStd::pair<AZStd::pair<uint32_t, uint32_t>, AZStd::string>> valuesToLabels =
+        {
+            { {1280, 720}, "1280 x 720" },
```
lmbr-pip, line 286: Wonder if its worth adding a note in the README about this list or adding new settings?

Gem/Code/Source/Components/UI/UiSettingsComponent.h
```diff
@@ -24,6 +25,7 @@ namespace MultiplayerSample
 
     class UiSettingsComponent
```
lmbr-pip, line 26: Would love a class on all public classes ;)

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -180,6 +188,99 @@ namespace MultiplayerSample
...
+                    &AzFramework::WindowRequestBus::Events::SetFullScreenState, fullscreen);
+            }
+
+            registry->Set(m_fullscreenKey.c_str(), fullscreen);
+        }
+    }
+
+    AZStd::pair<uint32_t, uint32_t> MultiplayerSampleUserSettings::GetResolution()
+    {
+        uint64_t width = 1920;
```
lmbr-pip, line 230: nit: Feel like these should be easier to find as defaults ie named defaults with reason for how set. Otherwise they are hidden inside code.

lmbr-pip (approved):

AMZN-alexpete (approved):

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: Just to be 100% clear, this particular list of resolutions doesn't appear anywhere currently. See screenshot. The toggle related to this code is currently disabled because changing resolutions is a bad UX experience right now.

I suppose we could add a note to the README about the settings screen though. Separate PR?

Gem/Code/Source/Components/UI/UiSettingsComponent.h
mbalfour-amzn: Added a comment, lmk if there's anything else you'd like it to cover.

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
lmbr-pip: Seperate PR is fine, if you have time.

lmbr-pip: Adding brief comment to code though will help readability as to why code does nothing ;)

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: I'm totally open to suggestions here. 

I could try and grab each value from its associated cvar, setreg setting, etc, though some things like master volume don't even have an exposed setting anywhere, and I think this class initializes too early to guarantee that it _could_ grab defaults from all of the associated places.

I could create a little const struct that contains all the defaults in it and place it at the top of the file.

I could try and read the defaults in from a data file or setreg file, but then that brings me back to the position of needing default values if those don't exist...

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: I intentionally kept the comment out of the code because it's purely a data change that determines if/when it becomes functional. It would be a little bit like putting a disclaimer in credits.md to say that there's no screen that loads and shows it yet. But happy to add the comment if you think it's less readable right now.

PR #391, opened 2023-04-28
Multiplayer Sample GameLift Exporter Script

AMZN-Gene:
Starting an exporter script for Multiplayer Sample GameLift. 
Removed some steps from the Readme
Note: Currently doesn't support Multiplayer Sample built with engine SDK

Future Improvements:
 - SDK engine support
 - Automatically export zip of exes, dlls, and pak files
 - Need automated testing to ensure this script doesn't break

### DIFF
--- a/MPSGameLift/Documentation/GameLift.md
+++ b/MPSGameLift/Documentation/GameLift.md
@@ -64,9 +64,9 @@ It's important to make sure that the bootstrap.game.profile.setreg file has been
 1. Test the profile pak server and game locally
     Run the server in headless mode using `rhi=null` and `NullRenderer` parameters; the server appears as a white screen in headless mode.
     
-    `C:\GameLiftPackageWindows\MultiplayerSample.ServerLauncher.exe --rhi=null -NullRenderer -bg_ConnectToAssetProcessor=0 -sys_PakPriority=2 -sv_terminateOnPlayerExit=true --console-command-file=launch_server.cfg`
+    `C:\GameLiftPackageWindows\MultiplayerSample.ServerLauncher.exe --rhi=null -NullRenderer -bg_ConnectToAssetProcessor=0 -sys_PakPriority=2 --console-command-file=launch_server.cfg`
     
-    `C:\GameLiftPackageWindows\MultiplayerSample.GameLauncher.exe -bg_ConnectToAssetProcessor=0 -sys_PakPriority=2 --connect`
+    `<path-to-multiplayer-sample>\build\windows\bin\profile\MultiplayerSample.GameLauncher.exe -bg_ConnectToAssetProcessor=0 --connect`
 
     ---
     **NOTE**
--- a/MPSGameLift/Documentation/GameLift.md
+++ b/MPSGameLift/Documentation/GameLift.md
@@ -23,6 +23,12 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
     ```sh
     <path-to-o3de-engine>\scripts\o3de.bat export-project -es <path-to-multiplayer-sample>\MPSGameLift\Scripts\export_gamelift_server_package.py --code --assets -ll INFO
     ```
+    ---
+    **Important**
+
+    The export_gamelift_server_package script only works for projects built using engine source, and won't work with engine as an sdk. 
+
+    ---
     
 1. Bundle Content
 
--- a/MPSGameLift/Documentation/GameLift.md
+++ b/MPSGameLift/Documentation/GameLift.md
@@ -30,14 +30,12 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
 
     ---
     
-1. Bundle Content
+    ---
+    **Important**
 
-    a. Open .\build\windows\bin\profile\AssetBundler.exe
+    The export_gamelift_server_package script only works for projects built using engine source, and won't work with engine as an sdk. 
 
-    b. Follow steps for "Create a bundle for game assets" and "Create a bundle for engine assets" and "Add bundles to the release game layout" here: https://www.o3de.org/docs/user-guide/packaging/asset-bundler/bundle-assets-for-release/
-
-The "default seed lists" choice should choose all but 4 seed lists to make the engine_pc.pak
-The other 4 seed lists should all get selected to make the game_pc.pak
+    ---
 It's important to make sure that the bootstrap.game.profile.setreg file has been added to one of the seed files. (also add debug if you want to support debug builds)
 
 1. Create the Launcher Zip file
--- a/MPSGameLift/Documentation/GameLift.md
+++ b/MPSGameLift/Documentation/GameLift.md
@@ -12,49 +12,18 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
     C:\> aws --version
     aws-cli/2.10.0 Python/3.11.2 Windows/10 exe/AMD64 prompt/off
     ```
-1. Enable the "AWSGameLift" and "MPSGameLift" gems
-
-    Option 1: Commandline
-    ```sh
-    <path-to-o3de-engine>\scripts\o3de.bat enable-gem -pp <path-to-multiplayer-sample> -gn MPSGameLift
-    ```
-    Option 2: Project Manager
-    1. Open Project Manager
-    2. Select the "Configure Gems" options for Multiplayer Sample
-    3. Enable "AWSGameLift" and "MPSGameLift" gems
-
-        ![Enable GameLift Gems](Media/enable_gamelift_gems.jpg)
-    4. Click "Save"
-
-
-    ---
-    **NOTE**
-
-    Due to [bug-15829](https://github.com/o3de/o3de/issues/15829) please also add "AWSGameLift" to project.json's _gem_names_ list.
-
-    ---
-    
-
-1. Build the server, game launchers, and asset bundler for MultiplayerSample
-
-    `cmake --build build\windows --target Editor MultiplayerSample.GameLauncher MultiplayerSample.ServerLauncher AssetBundler --config profile -- /m `
-
-1. Build all the assets
-
-    `cmake --build build\windows --target MultiplayerSample.Assets --config profile -- /m`
-
 1. Work in progress (WiP) step: Add your AWS region to Config/default_aws_resource_mappings.json (example: "Region": "us-west-2")
 
     a. Currently needed otherwise when the client initializes GameLift there will be an error about not having a region.
 
     b. This step will be removed once we properly parse the game-session data which contains the fleet-id, region-id, etc  
 
-## Build Server for Windows
-1. Build Monolithic Server
+1. Use Export Project to Compile Code and Build Assets
 
-    a. `cmake -B build\windows_mono -S . -G "Visual Studio 16" -DLY_MONOLITHIC_GAME=1 -DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0`
-
-    b. `cmake --build build\windows_mono --target MultiplayerSample.GameLauncher MultiplayerSample.ServerLauncher --config profile -- /m /nologo`
+    ```sh
+    <path-to-o3de-engine>\scripts\o3de.bat export-project -es <path-to-multiplayer-sample>\MPSGameLift\Scripts\export_gamelift_server_package.py --code --assets -ll INFO
+    ```
+    
 1. Bundle Content
 
     a. Open .\build\windows\bin\profile\AssetBundler.exe
@@ -64,6 +33,7 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
 The "default seed lists" choice should choose all but 4 seed lists to make the engine_pc.pak
 The other 4 seed lists should all get selected to make the game_pc.pak
 It's important to make sure that the bootstrap.game.profile.setreg file has been added to one of the seed files. (also add debug if you want to support debug builds)
+
 1. Create the Launcher Zip file
    Use the following .bat file or equivalent copy steps to create a directory with the launchers in it:
    Run from MultiplayerSample project root directory...
@@ -97,7 +67,7 @@ It's important to make sure that the bootstrap.game.profile.setreg file has been
     ---
     **NOTE**
 
-    Note: launch_server.cfg is required because there's a bug with multiplayer when calling --loadlevel in the command-line. See https://github.com/o3de/o3de/issues/15773.
+    Launch_server.cfg is required because there's a bug with multiplayer when calling --loadlevel in the command-line. See https://github.com/o3de/o3de/issues/15773.
 
     ---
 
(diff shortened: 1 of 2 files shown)


MPSGameLift/Documentation/GameLift.md
```diff
@@ -12,49 +12,18 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
...
-
 1. Work in progress (WiP) step: Add your AWS region to Config/default_aws_resource_mappings.json (example: "Region": "us-west-2")
 
     a. Currently needed otherwise when the client initializes GameLift there will be an error about not having a region.
 
     b. This step will be removed once we properly parse the game-session data which contains the fleet-id, region-id, etc  
 
-## Build Server for Windows
-1. Build Monolithic Server
+1. Use Export Project to Compile Code and Build Assets
```
tkothadev, line 21: Do we want to add a disclaimer here regarding using an SDK flag for those using o3de release?

MPSGameLift/Documentation/GameLift.md
```diff
@@ -64,6 +33,7 @@ This README covers optional setup, testing and running on [Amazon GameLift](http
 The "default seed lists" choice should choose all but 4 seed lists to make the engine_pc.pak
 The other 4 seed lists should all get selected to make the game_pc.pak
 It's important to make sure that the bootstrap.game.profile.setreg file has been added to one of the seed files. (also add debug if you want to support debug builds)
```
tkothadev, line 35: If the location for the setreg file is known, you can validate for this in the export script as well

MPSGameLift/Documentation/GameLift.md
```diff
@@ -97,7 +67,7 @@ It's important to make sure that the bootstrap.game.profile.setreg file has been
     ---
     **NOTE**
 
-    Note: launch_server.cfg is required because there's a bug with multiplayer when calling --loadlevel in the command-line. See https://github.com/o3de/o3de/issues/15773.
+    Launch_server.cfg is required because there's a bug with multiplayer when calling --loadlevel in the command-line. See https://github.com/o3de/o3de/issues/15773.
```
tkothadev, line 70: like with .setreg files, you can validate to confirm the file exists as you expect it.

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+                    prog='GameLift Server Package',
+                    description='Helps setup the project for AWS Gamelift and creates a Windows server package which can be uploaded to a GameLift.')
+
+parser.add_argument('--code', action='store_true', help='Build code')
+parser.add_argument('--assets', action='store_true', help='Build assets')
+
+args = parser.parse_args(o3de_context.args)
+
+if not args.code and not args.assets:
+    user_input = input('No build command specified. Do you want to build code, assets, or both? (c/a/b): ')
```
tkothadev, line 40: I'm thinking it'd be good to have this in a validation while loop, otherwise the variables are not set if the user doesn't press the correct key.

Also should we allow a `q` key, to quit if the user changes their mind?

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+    if (process_command(["cmake", "-B", monolithic_build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16", "-DLY_MONOLITHIC_GAME=1", "-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0"])):
+        quit()
+
+    if (process_command(["cmake", "--build", monolithic_build_folder, "--target", "MultiplayerSample.ServerLauncher", "--config", "profile", "--", "/m"])):
+        quit()
+
+# Build Assets
+if (args.assets):
+    # Process assets
+    if (process_command(["cmake", "--build", build_folder, "--target", "MultiplayerSample.Assets", "--config", "profile", "--", "/m"]) != 0):
```
tkothadev, line 87: Were there any CMake targets available for the asset bundler, or will that require a separate step?

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Build server launcher
+    build_folder = os.path.join(o3de_context.project_path, "build", "windows")
+    os.makedirs(build_folder, exist_ok=True)
+    o3de_logger.info(f"Building {project_name}.ServerLauncher")
+
+    if (process_command(["cmake", "-B", build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16"])):
+        quit()
+
+    if (process_command(["cmake", "--build", build_folder, "--target", f"{project_name}.ServerLauncher", "--config", "profile", "--", "/m"]) != 0):
```
tkothadev, line 69: nit: I'm noticing some if statements where you do:
`if (process_command(...) != 0)`
and some where you do:
`if (process_command(...))`

It'd be best to stick to one unless there's a strong reason for exceptions.

MPSGameLift/Scripts/export_gamelift_server_package.py
tkothadev: Saw that asset bundling is marked as a future improvement. I'll still keep my question around in case other contributors can chime in here.

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if (enable_gem.enable_gem_in_project(gem_name="MPSGameLift", project_path=o3de_context.project_path) != 0):
+        quit()
+
+    # Build server launcher
+    build_folder = os.path.join(o3de_context.project_path, "build", "windows")
+    os.makedirs(build_folder, exist_ok=True)
+    o3de_logger.info(f"Building {project_name}.ServerLauncher")
+
+    if (process_command(["cmake", "-B", build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16"])):
```
AMZN-alexpete, line 66:
```suggestion
    if (process_command(["cmake", "-B", build_folder, "-S", o3de_context.project_path])):
```
pretty sure if you don't specify `-G` CMake will default to the latest installed Visual Studio version.

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+        quit()
+
+    if (process_command(["cmake", "--build", build_folder, "--target", f"{project_name}.ServerLauncher", "--config", "profile", "--", "/m"]) != 0):
+        quit()
+        
+    if (process_command(["cmake", "--build", build_folder, "--target", f"AssetBundler", "--config", "profile", "--", "/m"]) != 0):
+        quit()
```
AMZN-alexpete, line 73: I'd just build them both at the same time 
```suggestion
    if (process_command(["cmake", "--build", build_folder, "--target", f"{project_name}.ServerLauncher", "AssetBundler", "--config", "profile", "--", "/m"]) != 0):
        quit()
```

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Build monolithic server launcher build
+    monolithic_build_folder = os.path.join(o3de_context.project_path, "build", "windows_mono")
+    os.makedirs(monolithic_build_folder, exist_ok=True)
+    if (process_command(["cmake", "-B", monolithic_build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16", "-DLY_MONOLITHIC_GAME=1", "-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0"])):
+        quit()
+
+    if (process_command(["cmake", "--build", monolithic_build_folder, "--target", f"{project_name}.ServerLauncher", "--config", "profile", "--", "/m"])):
+        quit()
```
AMZN-alexpete, line 82: does profile monolithic version work when using SDK? - and also recommend having a per-platform function in here or settings so you can run this on Linux too.
```suggestion
    monolithic_build_folder = os.path.join(o3de_context.project_path, "build", "windows_mono")
    os.makedirs(monolithic_build_folder, exist_ok=True)
    if (process_command(["cmake", "-B", monolithic_build_folder, "-S", o3de_context.project_path, "-DLY_MONOLITHIC_GAME=1", "-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0"])):
        quit()

    if (process_command(["cmake", "--build", monolithic_build_folder, "--target", f"{project_name}.ServerLauncher", "--config", "profile", "--", "/m"])):
        quit()
```

MPSGameLift/Documentation/GameLift.md
tkothadev: for now, something as simple as a file exists function can help

MPSGameLift/Documentation/GameLift.md
AMZN-Gene: Added an *important* note saying the scripting currently wont work with engine as sdk. I'll make a future pr

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Thanks! Updated!

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Good catch, fixed!

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Fixed!

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: fixed!

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Monolithic version work when using SDK only if the SDK was built using some proper flags. I'll need a future PR for building SDK. I've added a note that this script won't currently help if using engine as sdk.

MPSGameLift/Documentation/GameLift.md
AMZN-Gene: Definitely! I'll add the checks in a future PR when I actually create the "GameLiftServerPackageWindows" folder.

MPSGameLift/Documentation/GameLift.md
AMZN-Gene: Definitely! I can add checks in the next PR when I actually bundle the assets. Still not sure how we handle that, as a game may have any number of seedlists.

tkothadev (approved):

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+    if (process_command(["cmake", "-B", build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16"])):
+        quit()
+
+    if (process_command(["cmake", "--build", build_folder, "--target", f"{project_name}.ServerLauncher", "AssetBundler", "--config", "profile", "--", "/m"]) != 0):
+        quit()
+        
+    # Build monolithic server launcher build
+    monolithic_build_folder = os.path.join(o3de_context.project_path, "build", "windows_mono")
+    os.makedirs(monolithic_build_folder, exist_ok=True)
+    if (process_command(["cmake", "-B", monolithic_build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16", "-DLY_MONOLITHIC_GAME=1", "-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0"])):
```
cgalvan, line 78: It would be nice if we could pass in the generator in case the user wanted to use `Visual Studio 17` instead

MPSGameLift/Scripts/export_gamelift_server_package.py
```diff
@@ -0,0 +1,88 @@
...
+    o3de_logger.info(f"Building {project_name}.ServerLauncher")
+
+    if (process_command(["cmake", "-B", build_folder, "-S", o3de_context.project_path, "-G", "Visual Studio 16"])):
+        quit()
+
+    if (process_command(["cmake", "--build", build_folder, "--target", f"{project_name}.ServerLauncher", "AssetBundler", "--config", "profile", "--", "/m"]) != 0):
+        quit()
+        
+    # Build monolithic server launcher build
+    monolithic_build_folder = os.path.join(o3de_context.project_path, "build", "windows_mono")
```
cgalvan, line 76: Could we get this path from cmake presets somehow?

cgalvan (approved):

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Updated! Added asset bundling. This reads the game and engine's seedlists to create a full asset list, and then generates an asset bundle via the asset list.

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: Good thinking, updated!

MPSGameLift/Scripts/export_gamelift_server_package.py
AMZN-Gene: I see there's CMAKE_LIBRARY_OUTPUT_DIRECTORY (ex value: C:/o3de-multiplayersample/build/windows/bin) for normal builds, but no presets for a monolithic build folder. And even for the normal builds, that provides the output directory for the binaries, and not the build directory per-se. It's 100% possible I'm missing something though.

tkothadev:
regarding automated testing:
perhaps we should treat these export scripts as the _mechanism_ for automated testing?

The way I think we can phrase it is like: "If an export script breaks, that means there was something about the engine that changed from the last execution of this script."

AMZN-Gene:
> regarding automated testing: perhaps we should treat these export scripts as the _mechanism_ for automated testing?
> 
> The way I think we can phrase it is like: "If an export script breaks, that means there was something about the engine that changed from the last execution of this script."

I like that idea. Export Automated Testing. Not sure what part of AR it would fit into... if AR already builds assets and game/server launcher, I wonder how long the export would take? Fast enough to add it to AR? Maybe needs to be part of just the nightly run.

tkothadev:
If resources are strained for testing exports, we may want to reserve it for closer to release cycles. That way we can be sure we're preventing bugs from being introduced to new releases.

amzn-changml:
Thanks for the suggestion! This seems reasonable and I mostly agree with what's being proposed. In an ideal world, we would automatically test every phase of the engine workflow, from build, project creation, project export end to end in one go within AR. We do want to cap our pipeline execution time (due to infra costs/velocity), however, so there's limits to what we can to do within our time budget. 

That said, running it in our periodic builds sounds like a good start; we already test some basic things in our daily installer builds (download, install, start engine), so it's not too much of a stretch to add project creation/export as a separate pipeline step. Ideally, we want to use a representative test project that covers a good portion of use cases, whether it be AutomatedTest or some other test project.

Feel free to make a GHI to SIG-Build with your ideas and we can bring it to TSC to further discussion. We can prioritize based on this.

PR #393, opened 2023-05-02
Add a resolution setting to the settings screen

mbalfour-amzn:
This adds a resolution setting to the MPS settings screen. 

* It currently only affects the resolution while in Windowed mode, Atom doesn't yet support changing the resolution while in Fullscreen mode.
* It will only offer resolutions that fit on the current screen.
* The settings screen itself has been modified to resize correctly for different aspect ratios.

Also fixed a small problem in NewStarbase by removing a bad override patch that threw warnings every time the level was opened.

image


Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -258,40 +258,83 @@ namespace MultiplayerSample
...
 
         // Rotate the index based on toggle direction.
-        uint32_t fullscreenIndex = GetRotatedIndex(valuesToLabels, fullscreen, toggleDirection);
+        uint32_t fullscreenIndex = GetRotatedIndex(valuesToLabels, currentFullscreenState, toggleDirection);
 
         UiTextBus::Event(toggle.m_labelEntity, &UiTextInterface::SetText, valuesToLabels[fullscreenIndex].second);
 
-        MultiplayerSampleUserSettingsRequestBus::Broadcast(
-            &MultiplayerSampleUserSettingsRequestBus::Events::SetFullscreen, valuesToLabels[fullscreenIndex].first);
+        // Only set & save the state if the value has changed. We might get a lot of spurious notifications here, so
```
lmbr-pip, line 277: nit: if we no why we get supurious notifications may be good to add ie on Windows the OS appears to send this multiple times..

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -258,40 +258,83 @@ namespace MultiplayerSample
         // Get the current fullscreen state. Unlike the other settings, we'll get this from the current window state so that we
         // handle things like Alt-enter that can change our windowing state regardless of what our user settings thinks.
 
-        // Start by defaulting to the user setting.
+        // Start by getting the user setting as our default state.
```
lmbr-pip, line 261: nit: avoid personal pronouns in comment, its not clear to readers who "ours" is, o3de? the sample?

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -22,6 +25,15 @@
 
 namespace MultiplayerSample
 {
+    static constexpr const char* DefaultGraphicsApi = "";       // default to the platform-specific default graphics API
```
lmbr-pip, line 28: FTW!

mbalfour-amzn: Thanks for suggesting this on a previous PR! :)

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: Actually this comment is probably no longer true, and I probably don't need the extra safeguards. This was from a previous iteration where I hadn't added the OnFullscreenModeChanged() notification yet, so I was having to listen to other notifications that fired more often. I'll see if I can just remove this bit.

Gem/Code/Source/Components/UI/UiSettingsComponent.h
```diff
@@ -40,8 +40,7 @@ namespace MultiplayerSample
         void Deactivate() override;
     private:
         // WindowNotificationBus overrides
-        void OnWindowResized(uint32_t width, uint32_t height) override;
```
moudgils, line 43: Is OnWindowResized not needed?

moudgils (approved):

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
mbalfour-amzn: Yep, it was safe to remove. :)

Gem/Code/Source/Components/UI/UiSettingsComponent.cpp
```diff
@@ -258,40 +258,83 @@ namespace MultiplayerSample
...
+                &MultiplayerSampleUserSettingsRequestBus::Events::SetFullscreen, valuesToLabels[fullscreenIndex].first);
 
-        MultiplayerSampleUserSettingsRequestBus::Broadcast(&MultiplayerSampleUserSettingsRequestBus::Events::Save);
+            MultiplayerSampleUserSettingsRequestBus::Broadcast(&MultiplayerSampleUserSettingsRequestBus::Events::Save);
+        }
     }
 
     void UiSettingsComponent::OnResolutionToggle(UiToggle& toggle, ToggleDirection toggleDirection)
     {
         const AZStd::vector<AZStd::pair<AZStd::pair<uint32_t, uint32_t>, AZStd::string>> valuesToLabels =
```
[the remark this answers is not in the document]
mbalfour-amzn: Thanks, this is much cleaner and nicer!

Gem/Code/Source/Components/UI/UiSettingsComponent.h
mbalfour-amzn: Correct. It was previously being used as a poor/indirect way to detect fullscreen changes, though it only actually worked if going to fullscreen changed the size of the window. 

I've added an OnFullScreenModeChanged() notification that we're now using instead, which provides a reliable way to detect the state change.

lumberyard-employee-dm (approved):

PR #400, opened 2023-05-10
MPSGameLift: Adding GameLift Runtime Dependencies

AMZN-Gene:
Adding GameLift runtime dependencies for MPSGameLift so that the GameLift system components are added.

Fixes https://github.com/o3de/o3de/issues/15829

Tested by adding MPSGameLift to project.json gem dependencies, and then running the gamelauncher, serverlauncher, and unifiedlauncher and seeing everything work as normal (not crash)

### DIFF
--- a/MPSGameLift/Code/CMakeLists.txt
+++ b/MPSGameLift/Code/CMakeLists.txt
@@ -64,7 +64,6 @@ ly_add_target(
             Gem::AWSCore.Static
             Gem::AWSGameLift.Client.Static
             Gem::LyShine.Static
-
 )
 
 ly_add_target(
@@ -137,7 +136,6 @@ ly_add_target(
             Gem::AWSCore.Static
             Gem::AWSGameLift.Client.Static
             Gem::${gem_name}.Client.Private.Object
-
     RUNTIME_DEPENDENCIES
         Gem::AWSGameLift.Clients
 )
@@ -159,7 +157,6 @@ ly_add_target(
         PRIVATE
             Gem::Multiplayer.Server.Static
             Gem::${gem_name}.Server.Private.Object
-
     RUNTIME_DEPENDENCIES
         Gem::AWSGameLift.Servers
 )
--- a/MPSGameLift/Code/CMakeLists.txt
+++ b/MPSGameLift/Code/CMakeLists.txt
@@ -137,6 +137,9 @@ ly_add_target(
             Gem::AWSCore.Static
             Gem::AWSGameLift.Client.Static
             Gem::${gem_name}.Client.Private.Object
+
+    RUNTIME_DEPENDENCIES
+        Gem::AWSGameLift.Clients
 )
 
 ly_add_target(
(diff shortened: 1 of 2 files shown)


MPSGameLift/Code/CMakeLists.txt
```diff
@@ -137,6 +137,9 @@ ly_add_target(
             Gem::AWSCore.Static
             Gem::AWSGameLift.Client.Static
             Gem::${gem_name}.Client.Private.Object
+
```
AMZN-alexpete, line 140: extra blank newline you rebel you 😄

AMZN-alexpete (approved):
@gene-walt  Can you verify the `Unified` launcher works? you may need to add `Gem::AWSGamelift` to `TARGETS` on line 190, or add `RUNTIME_DEPENDENCIES` on line 183 for the `Gem` target? @lumberyard-employee-dm are the `Unified` `RUNTIME_DEPENDENCIES` basically the `Gem` target in which case `Gem::AWSGamelift` could be added to that target's `RUNTIME_DEPENDENCIES`?

MPSGameLift/Code/CMakeLists.txt
AMZN-Gene: Sharp eye. Fixed!

lumberyard-employee-dm:
> @gene-walt Can you verify the `Unified` launcher works? you may need to add `Gem::AWSGamelift` to `TARGETS` on line 190, or add `RUNTIME_DEPENDENCIES` on line 183 for the `Gem` target? @lumberyard-employee-dm are the `Unified` `RUNTIME_DEPENDENCIES` basically the `Gem` target in which case `Gem::AWSGamelift` could be added to that target's `RUNTIME_DEPENDENCIES`?

Since the Unified launcher references the `${gem_name}` TARGET which corresponds to the `MPSGameLift` MODUJLE target, that is where the RUNTIME_DEPENDENCIES should be added.

RUNTIME_DEPENDENCIES should only be added to ly_target calls that declare themselves as a gem target by specifying `GEM_MODULE` or `${PAL_TRAIT_MONOLITHIC_DRIVEN_MODULE TYPE}`
Alternatively the dependencies can be added as additional targets to the ly_create_alias call for the `${gem_name}.Unified` target.

In this case the raw `${gem_name}` gem MODULE target doesn't seem to have a real load dependency on the `Gem::AWSGameLift` and it is only when the launcher is used in a "unified" manner through the `${gem_name}.Unified` target that it needs to load the dependent gem

AMZN-Gene:
> @gene-walt Can you verify the `Unified` launcher works? you may need to add `Gem::AWSGamelift` to `TARGETS` on line 190, or add `RUNTIME_DEPENDENCIES` on line 183 for the `Gem` target? @lumberyard-employee-dm are the `Unified` `RUNTIME_DEPENDENCIES` basically the `Gem` target in which case `Gem::AWSGamelift` could be added to that target's `RUNTIME_DEPENDENCIES`?

Unified launcher doesn't crash.

lumberyard-employee-dm (approved):

cgalvan (approved):

PR #399, opened 2023-05-09
Enhanced packaging doc guide

AMZN-Olex:
First rough draft update to the packaging doc guide.

The primary change is making it clear that one needs to use the source engine, not the installer and that one needs to produce both non-monolithic and monolithic engines and project binaries.

PR #401, opened 2023-05-10
Reverting README.md to clarify optional steps

mcphedar:
Reverting the README back to its original state. In Step 2, the steps are optional. If you register (option 1), you can then proceed to step 3. OR you can choose to use the Project Manager option, which will register the engine and gems and create the build, allowing you to skip step 3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ If you have problems with working with LFS, see the troubleshooting guide: https
 
 ## Step 2. Register the engine, the project, and the Gems
 
-### Required step to compile
+### Option #1 - Use the CLI
 
 <span style="background-color:#4F3C3C">**NOTE:** The following steps only need to be performed once.</span>
 
@@ -166,7 +166,7 @@ If you have problems with working with LFS, see the troubleshooting guide: https
 
 The final step prints warnings that the compatibility check for MultiplayerSample and Blast will be skipped. These warnings can be ignored.
 
-### Build Using the Project Manager
+### Option #2 - Use Project Manager
 
 If you've already built the O3DE engine, use the O3DE project manager to open an existing project.
 


AMZN-Olex:
The enhancement doc on this is in progress here: https://github.com/o3de/o3de-multiplayersample/pull/399

AMZN-Olex (approved):

README.md
```diff
@@ -142,7 +142,7 @@ If you have problems with working with LFS, see the troubleshooting guide: https
 
 ## Step 2. Register the engine, the project, and the Gems
 
-### Required step to compile
+### Option #1 - Use the CLI
```
lmbr-pip, line 145:
```suggestion
### Use the CLI to register engine and gems
```

README.md
```diff
@@ -166,7 +166,7 @@ If you have problems with working with LFS, see the troubleshooting guide: https
 
 The final step prints warnings that the compatibility check for MultiplayerSample and Blast will be skipped. These warnings can be ignored.
 
-### Build Using the Project Manager
+### Option #2 - Use Project Manager
```
lmbr-pip, line 169:
```suggestion
### (Optional) - Configure via Project Manager
```

OBWANDO (changes requested):
This is still incorrect. I spent 30 minutes with Games from scratch solving this and he literally used 30 seconds in his video to clarify that to people that they must register the gems because it did not work with the manager and fails when you skip that step thinking it's optional.

AMZN-alexpete:
It's very possible those gem registration issues - especially the PopcornFX ones will be resolved in dev with this PR https://github.com/o3de/o3de/pull/15933 unfortunately, that won't help anyone using the 2305 installer

OBWANDO:
It would be better to adjust the document than to revert it back into an incorrect state where people will become frustrated and lose faith in the project.

PR #403, opened 2023-05-10
Move music to the MX bus to be controlled by Volume_Music.

mbalfour-amzn:
Previously, the music tracks were on the SX bus, so sound effects and music were both being controlled by Volume_SFX. By moving the music to the MX bus, we can control its volume separately via Volume_Music.

This has been tested with upcoming changes to the Settings screen that offers separate volume controls for Master, Music, and SFX.


AMZN-Olex (approved):

kberg-amzn (approved):
Very good change.. something I was unable to get working with my limited knowledge of wwise. Thanks!

PR #405, opened 2023-05-11
Add more settings to the settings screen

mbalfour-amzn:
This changes the settings screen to have 3 tabs - General, Graphics, and Audio. The following settings have been added or changed:
* Graphics API - now filters out "null" and only shows the choices available on that platform.
* Reflection - turns SSR on and off
* MSAA - 1x, 2x, 4x (requires manual restart, Atom will potentially crash if this gets changed at runtime)
* TAA - on/off
* Music Volume
* Sfx Volume

I also needed to re-save a bunch of Script Canvas files because I modified the Script Events asset that they relied on to add events for changing between the 3 tabs.

image
imageimage


Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -26,21 +30,35 @@
...
+    static constexpr bool DefaultTaa = true;                    // default to TAA enabled
 
 
     MultiplayerSampleUserSettings::MultiplayerSampleUserSettings()
         : m_graphicsApiKey(BaseRegistryKey + FixedString("/ApiName"))
-        , m_masterVolumeKey(BaseRegistryKey + FixedString("/MasterVolume"))
+        , m_volumeKey{
+            BaseRegistryKey + FixedString("/MasterVolume"),
+            BaseRegistryKey + FixedString("/MusicVolume"),
+            BaseRegistryKey + FixedString("/SfxVolume"),
```
lumberyard-employee-dm, line 53: There is a bit of a trick that can done using the call operator to have the BaseRegistryKey field return new fixed_string.

It is used in the event logger to encapsulate appending of keys to the root metrics key [here](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Metrics/IEventLogger.h#L30-L74).

It makes it possible to return a key anchored at the specific field by using the call operator such  `BaseRegistryKey("MasterVolume")`

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -200,6 +247,135 @@ namespace MultiplayerSample
...
+        return static_cast<SpecularReflections>(reflectionType);
+    }
+
+    void MultiplayerSampleUserSettings::SetReflectionSetting(SpecularReflections reflectionType)
+    {
+        if (auto* registry = AZ::SettingsRegistry::Get(); registry != nullptr)
+        {
+            // Only try to set the settings if the scene system is active.
+            // If we try to set it too early, it will assert / crash.
+            if (AzFramework::SceneSystemInterface::Get())
```
AMZN-alexpete, line 268: wonder if it would be simpler to make a `m_registry` member and only set it once the dependent interfaces are ready and then replace all the `registry = AZ::SettingsRegistry::Get()` calls with `if(m_registry)`

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.h
```diff
@@ -9,9 +9,32 @@
...
+    };
+
+    enum SpecularReflections : uint8_t
+    {
+        None,
+        ScreenSpace,
+        ScreenSpaceAndRaytracing
+    };
+
+    enum Msaa : uint8_t
```
lumberyard-employee-dm, line 31: minor: These enums can be made `enum class`

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
AMZN-alexpete: could potentially wait to call `MultiplayerSampleUserSettingsRequestBus` till m_registry is ready and then remove the `if(m_registry)` conditionals

AMZN-alexpete (approved):
just some minor suggestions, also a request - can you add a 'Auto head-shot' toggle and 'Invincible' toggle to the settings that do nothing 😄 ?  Then we can wait for the GHIs "Auto head-shot not working"

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.h
mbalfour-amzn: I specifically made VolumeChannel _not_ class so that I could use that `Max` to declare a static array size without having to cast it. I can change the other two though.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
kberg-amzn: Always good to be as defensive as possible, particularly in non-performance sensitive use-cases like here. I'd prefer it if it always checked just to make sure, because that ensures the program would keep trucking along even if in a bad or inconsistent state.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.h
kberg-amzn: You could still do enum class with a using for max if you wanted:
`enum class VolumeChannel : uint8_t
{
   Master, 
   Music, 
   Sound
}
using VolumeChannelMax = static_cast<size_t>(VolumeChannel::Sound + 1);`

kberg-amzn (approved):

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: It's a neat trick, though for something like this I'd prefer to go more like the AZ Path route and use the / operator to join them so that the separators are more obvious. I'd also rather have a single helper class somewhere in AzCore/Settings that implements this than create little one-offs in multiple places. I'll leave this as-is for now, except for making them static constexpr strings per your other suggestion.

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: ok, so i'm a big fan of defensiveness, but i'll agree with @AMZN-alexpete that i went a little too far here, and that it's reasonable to cache the registry pointer in the constructor because the settings registry *needs* to be one of the first things started up for the engine to work correctly. It's actually a little _too_ defensive as written, because the settings would just silently fail to work correctly if the settings registry wasn't initialized. Now it will assert and crash, which is reasonable, because it would be a programming problem if this module started up in the wrong order.

The other defensive checks in here are still needed though because the settings can either get applied during boot or from the settings screen. Different subsets of systems are available depending on what point those settings calls occur.

AMZN-alexpete (approved):

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
```diff
@@ -55,159 +86,268 @@ namespace MultiplayerSample
...
     {
-        bool fullscreen = DefaultFullscreenMode;
+        // Currently MSAA settings don't get changed at runtime because they have the potential
+        // to cause a TDR graphics driver crash on at least Vulkan, maybe others.
+        // This might be the result of mixing PSOs between the msaa variant and the non-msaa variant in the same frame.
+        // Until the problem gets tracked down and resolved, we'll only set the MSAA setting in the renderer at boot time.
+        // If this ever gets fixed, the call to SetMsaaInRenderer() can get moved to here, and the extra flow handling to call
+        // it sooner can get removed.
+
+        m_registry->Set(MsaaKey.c_str(), aznumeric_cast<AZ::u64>(msaa));
```
moudgils, line 311: Switching between 2x to 4x should work fine. Only 1x to non-1x should require a restart. Although this is also fine for now.

moudgils (approved):

Gem/Code/Source/UserSettings/MultiplayerSampleUserSettings.cpp
mbalfour-amzn: Yeah, I deliberately chose to make them all work the same way. Otherwise, you end up with a Settings toggle that only works immediately 2/3 of the time and requires a restart 1/3 of the time, which just looks broken.

PR #416, opened 2023-05-24
Fix MPS energy balls

mbalfour-amzn:
The previous implementation of the MPS energy balls caused some odd artifacts on the client. Timing discrepancies could cause the energy balls to draw outside of their expected path, generate a bunch of warnings in the logs, and cause a bunch of explosions to happen at (0, 0, 0). 

This refactor relies on the changes in https://github.com/o3de/o3de/pull/16013 to guarantee property correctness at entity deletion.

Previously, the energy ball lifetime was controlled by the server as "start to end plus N seconds to allow for an explosion". The explosion command was sent as an RPC. However, if the RPC arrived after the N seconds, no explosion would happen. If the start and end both showed up at the same time due to network delays, the explosion would happen at (0, 0, 0). If the N-second end delete was delayed, the ball could keep travelling on the client beyond where the explosion occurred. (In part, this was because the server also continued to move the ball during those N seconds after the explosion)

With these changes, the lifetime is simply "start to end", and at the end, the client triggers a fire-and-forget explosion effect and applies the blast decals. It knows how to apply the blast because the HitEvent networked property is guaranteed to contain the hit information when the entity is deleted.

To support the fire-and-forget effect, GameEffect has been extended to allow that as a new optional effect type.

For the before and after videos, I turned on energy ball debug drawing with a 2-second trail to make it really easy to see the issues.

Before:
Here's are the artifacts to look at in the before video:
image

https://github.com/o3de/o3de-multiplayersample/assets/82224783/965095fa-2c31-4cee-81ae-90e6aa4fbe0a

After:
No extra explosions, balls always stop exactly where they explode at.
https://github.com/o3de/o3de-multiplayersample/assets/82224783/f32c6011-6044-4f4b-9783-ab242af4e44e


Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
```diff
@@ -39,66 +40,41 @@ namespace MultiplayerSample
...
-        m_ballActiveHandler.Disconnect();
+        m_effect = {};
+        AZ::EntityBus::Handler::BusDisconnect();
+        m_debugDrawEvent.RemoveFromQueue();
 #endif
     }
 
 #if AZ_TRAIT_CLIENT
-    void EnergyBallComponent::OnBallActiveChanged(bool active)
+    void EnergyBallComponent::OnEntityDeactivated([[maybe_unused]] const AZ::EntityId& entityId)
```
kberg-amzn, line 65: So this is essentially a 'pre deactivate' event that gets triggered on the entity bus for each entity that's about to deactivate?

Gem/Code/Source/Effects/GameEffect.cpp
```diff
@@ -44,41 +44,79 @@ namespace MultiplayerSample
...
+        m_popcornFx = nullptr;
+        m_audioSystem = nullptr;
+        m_emitter = nullptr;
+        m_audioProxy = nullptr;
+        m_audioTriggerId = INVALID_AUDIO_CONTROL_ID;
 #endif
     }
 
-    void GameEffect::Initialize()
+    void GameEffect::Initialize([[maybe_unused]] EmitterType emitterType)
```
kberg-amzn, line 95: With these changes I'm a little nervous about someone calling initialize more than once.. It might be better to take the emitterType param at construct time and never allow it to be changed, and then make init safe to call as many times as we want, just subsequent calls can be no-ops I guess..?

AMZN-alexpete (approved):

Gem/Code/Source/Components/Multiplayer/EnergyBallComponent.cpp
mbalfour-amzn: Yep, exactly. You can see the flow here:
```
    void Entity::Deactivate()
    {
        AZ_PROFILE_FUNCTION(AzCore);

        AZ::ComponentApplicationRequests* componentApplication = AZ::Interface<AZ::ComponentApplicationRequests>::Get();
        if (componentApplication != nullptr)
        {
            componentApplication->SignalEntityDeactivated(this);
        }
        EntityBus::Event(m_id, &EntityBus::Events::OnEntityDeactivated, m_id);
        EntitySystemBus::Broadcast(&EntitySystemBus::Events::OnEntityDeactivated, m_id);

        AZ_Assert(m_state == State::Active, "Component should be in Active state to be Deactivated!");
        SetState(State::Deactivating);

        for (ComponentArrayType::reverse_iterator it = m_components.rbegin(); it != m_components.rend(); ++it)
        {
            DeactivateComponent(**it);
        }

        m_transform = nullptr;
        SetState(State::Init);
    }
```

A couple different types of "OnEntityDeactivated" events get triggered (one on a per-ID bus, one on a global bus), then the entity is put in a deactivating state, then the components are deactivated, then the entity is back in its init state.

OnEntityDeactivated is the last chance where you can process something while the entity and its components are still considered active.

Gem/Code/Source/Effects/GameEffect.cpp
mbalfour-amzn: Hmmm, maybe I don't understand the intent of the API here. I was thinking Initialize/Destroy were typically called in pairs roughly for construction/destruction, and then TriggerEffect/StopEffect were the main APIs that's called a bunch?

As long as you call Initialize/Destroy in pairs, it should work fine, and you can change the emitter type on each call to Initialize. I'll an assert and guards to catch double-initialization. I don't see any particular reason the emitter type _needs_ to go to the constructor? If we put it there, I'd be more inclined to make "IGameEffect" and have one shots and reusable be different subclasses. I considered this when adding it, but it seems like overkill to me.

PR #417, opened 2023-05-25
PopcornFX assets: Cleanup graph locations

HugoPKFX:
This PR does not touch effects behavior, only improves user experience when opening an effect.
Opening an effect (for most of them or a layer), we did not see any nodes. Had to pan the view to find where they are.

Captured nodegraph view location with bound nodes (did that for all effects under `Particles/` folder:
oxaf9eR4dn

Note: effects were opened & saved with PopcornFX Editor 2.16.3 (project was already upgraded to 2.16.2), this has no effect on the gem (effects will be re-baked without issues without needing a gem upgrade).


mbalfour-amzn:
You'll need to fix the DCO signing on your commits.

HugoPKFX:
Hi @mbalfour-amzn,

Sorry, re-created a different PR with the commit correctly signed this time: #421

PR #433, opened 2023-06-09
Created new MPS standalone export script for windows

tkothadev:
### Summary
This script automates building binaries, processing and bundling assets, and preparing a release directory layout for MPS. Currently it has been designed and tested for windows. A Linux compatible update will be for a future PR.

The script closely matches the instructions described here: https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md

The arguments for this script are as follows:
```
 --project-path
   required=True, Path to the intended O3DE project.

 --engine-path
   required=True, Path to the intended O3DE engine copy.

 --output-path
   required=True, Path that describes the final resulting Release Directory path location.

 --config
   default='profile', choices=['release', 'profile'], The CMake build configuration to use when building project binaries.

 --log-level
   default='INFO', choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'], Set the log level

 --build-non-mono-tools
   action='store_true', This option enables building of engine executables before building project.

 --non-mono-build-path
   default=None, Build path folder for non-monolithic binaries, like engine executables

 --mono-build-path
   default=None, Build path folder for monolithic binaries, like project executables. These are portable.

 --zip-output
   action='store_true', This option places the final output of the build into a zipped archive
```
An example execution can be found here:
```
# > <O3DE_ENGINE_ROOT_PATH>\python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -ps <O3DE_MPS_PROJECT_ROOT_PATH> -egn <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -zip
```
### Notes:
* We currently use an engine-centric approach for building executables.
* The script is designed to be standalone, so you can invoke it using either O3DE bundled python, or one locally on the machine. We currently duplicate project export functionality in this script, only because that feature is not available from the main branch of O3DE engine. Once it is available, this script will be revisited.

### Testing:
This script was tested against the main branches of o3de, o3de-multiplayersample, and o3de-multiplayersample-assets. I ran it multiple times, and verified that the server and game launcher executables worked correctly, per the instructions found here: https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md#run-the-final-build-to-verify-it

I also checked the asset bundle sizes for `engine_pc.pak` and `game_pc.pak`, and they matched the expected size limits of 72 MB and 1046 MB respectively.

### DIFF
--- a/ExportScripts/export_standalone_monolithic_windows.py
+++ b/ExportScripts/export_standalone_monolithic_windows.py
@@ -290,5 +290,5 @@ if __name__ == "__main__":
 
     logger.info(f"Exporting project is complete! Release Directory can be found at {args.output_path}")
 
-    if args.quiet:
+    if not args.quiet:
         process_command(['explorer', args.output_path])
--- a/ExportScripts/export_standalone_monolithic_windows.py
+++ b/ExportScripts/export_standalone_monolithic_windows.py
@@ -1,0 +1,285 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of https://www.github.com/o3de/o3de.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+import argparse
+import pathlib
+import logging
+import os
+import time
+import sys
+import subprocess
+import json
+
+from queue import Queue, Empty
+from threading  import Thread
+from typing import List
+from subprocess import Popen, PIPE
+
+logger = logging.getLogger('o3de.gamejam')
+LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
+#This is an export script for MPS on the Windows platform
+# this has to be a complete standalone script, b/c project export doesnt exist in main branch yet
+
+#View the argparse parameters for options available. An example invocation:
+
+# @<O3DE_ENGINE_ROOT_PATH>
+# > python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -ps <O3DE_MPS_PROJECT_ROOT_PATH> -egn <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -zip
+
+def enqueue_output(out, queue):
+    for line in iter(out.readline, b''):
+        queue.put(line)
+    out.close()
+
+def safe_kill_processes(*processes: List[Popen], process_logger: logging.Logger = None) -> None:
+    """
+    Kills a given process without raising an error
+    :param processes: An iterable of processes to kill
+    :param process_logger: (Optional) logger to use
+    """
+    def on_terminate(proc) -> None:
+        try:
+            process_logger.info(f"process '{proc.args[0]}' with PID({proc.pid}) terminated with exit code {proc.returncode}")
+        except Exception:  # purposefully broad
+            process_logger.error("Exception encountered with termination request, with stacktrace:", exc_info=True)
+
+    if not process_logger:
+        process_logger = logger
+    
+    for proc in processes:
+        try:
+            process_logger.info(f"Terminating process '{proc.args[0]}' with PID({proc.pid})")
+            proc.kill()
+        except Exception:  # purposefully broad
+            process_logger.error("Unexpected exception ignored while terminating process, with stacktrace:", exc_info=True)
+    try:
+        for proc in processes:
+            proc.wait(timeout=30)
+            on_terminate(proc)
+    except Exception:  # purposefully broad
+        process_logger.error("Unexpected exception while waiting for processes to terminate, with stacktrace:", exc_info=True)
+
+class CLICommand(object):
+    """
+    CLICommand is an interface for storing CLI commands as list of string arguments to run later in a script.
+    A current working directory, pre-existing OS environment, and desired logger can also be specified.
+    To execute a command, use the run() function.
+    This class is responsible for starting a new process, polling it for updates and logging, and safely terminating it.
+    """
+    def __init__(self, 
+                args: list,
+                cwd: pathlib.Path,
+                logger: logging.Logger,
+                env: os._Environ=None) -> None:
+        self.args = args
+        self.cwd = cwd
+        self.env = env
+        self.logger = logger
+        self._stdout_lines = []
+        self._stderr_lines = []
+    
+    @property
+    def stdout_lines(self) -> List[str]:
+        """The result of stdout, separated by newlines."""
+        return self._stdout_lines
+
+    @property
+    def stdout(self) -> str:
+        """The result of stdout, as a single string."""
+        return "\n".join(self._stdout_lines)
+
+    @property
+    def stderr_lines(self) -> List[str]:
+        """The result of stderr, separated by newlines."""
+        return self._stderr_lines
+
+    @property
+    def stderr(self) -> str:
+        """The result of stderr, as a single string."""
+        return "\n".join(self._stderr_lines)
+
+    def _poll_process(self, process, queue) -> None:
+        # while process is not done, read any log lines coming from subprocess
+        while process.poll() is None:
+            #handle readline in a non-blocking manner
+            try:  line = queue.get_nowait() 
+            except Empty:
+                pass
+            else: # got line
+                if not line: break
+                log_line = line.decode('utf-8', 'ignore')
+                self._stdout_lines.append(log_line)
+                self.logger.info(log_line)
+    
+    def _cleanup_process(self, process, queue) -> str:
+        # flush remaining log lines
+        while not queue.empty():
+            try: line = queue.get_nowait()
+            except Empty:
+                pass
+            else:
+                if not line: break
+                log_line = line.decode('utf-8', 'ignore')
+                self._stdout_lines.append(log_line)
+                self.logger.info(log_line)
+        stderr = process.stderr.read()
+
+        safe_kill_processes(process, process_logger = self.logger)
+
+        return stderr
+    
+    def run(self) -> int:
+        """
+        Takes the arguments specified during CLICommand initialization, and opens a new subprocess to handle it.
+        This function automatically manages polling the process for logs, error reporting, and safely cleaning up the process afterwards.
+        :return return code on success or failure 
+        """
+        ret = 1
+        try:
+            with Popen(self.args, cwd=self.cwd, env=self.env, stdout=PIPE, stderr=PIPE) as process:
+                self.logger.info(f"Running process '{self.args[0]}' with PID({process.pid}): {self.args}")
+
+                q = Queue()
+                t = Thread(target=enqueue_output, args=(process.stdout, q))
+                t.daemon = True
+                t.start()
+
+                process.stdout.flush()
+                self._poll_process(process, q)
+                stderr = self._cleanup_process(process, q)
+
+                ret = process.returncode
+
+                # print out errors if there are any      
+                if stderr:
+                    # bool(ret) --> if the process returns a FAILURE code (>0)
+                    logger_func = self.logger.error if bool(ret) else self.logger.warning
+                    err_txt = stderr.decode('utf-8', 'ignore')
+                    logger_func(err_txt)
+                    self._stderr_lines = err_txt.split("\n")
+        except Exception as err:
+            self.logger.error(err)
+            raise err
+        return ret
+
+# Helper API
+def process_command(args: list,
+                    cwd: pathlib.Path = None,
+                    env: os._Environ = None) -> int:
+    """
+    Wrapper for subprocess.Popen, which handles polling the process for logs, reacting to failure, and cleaning up the process.
+    :param args: A list of space separated strings which build up the entire command to run. Similar to the command list of subprocess.Popen
+    :param cwd: (Optional) The desired current working directory of the command. Useful for commands which require a differing starting environment.
+    :param env: (Optional) Environment to use when processing this command.
+    :return the exit code of the program that is run or 1 if no arguments were supplied
+    """
+    if len(args) == 0:
+        logging.error("function `process_command` must be supplied a non-empty list of arguments")
+        return 1
+    return CLICommand(args, cwd, logging.getLogger(), env=env).run()
+
+
+# EXPORT SCRIPT STARTS HERE!
+
+
+if __name__ == "__main__":
+    
+
+    parser = argparse.ArgumentParser(prog='Exporter for MultiplayerSample on windows',
+                                 description = "Exports O3DE's MultiplayerSample to the desired install directory in project layout...")
+    parser.add_argument('-ps', '--project-path', type=pathlib.Path, required=True, help='Path to the intended O3DE project.')
+    parser.add_argument('-egn', '--engine-path', type=pathlib.Path, required=True, help='Path to the intended O3DE engine copy.')
+    parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+    parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'], help='The CMake build configuration to use when building project binaries.')
+    parser.add_argument('-ll', '--log-level', default='INFO',
+                        choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
+                        help="Set the log level")
+    parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+    parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+    parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
+    parser.add_argument('-zip', '--zip-output', action='store_true', help='This option places the final output of the build into a zipped archive')
+    args = parser.parse_args()
+
+    logging.getLogger().setLevel(args.log_level)
+
+    non_mono_build_path = (args.engine_path) / 'build' / 'non_mono' if args.non_mono_build_path is None else args.non_mono_build_path
+    mono_build_path = (args.engine_path) / 'build' / 'mono' if args.mono_build_path is None else args.mono_build_path
+
+    #validation
+    assert args.engine_path.is_dir() and (args.engine_path / 'engine.json').is_file(), "Invalid Engine path provided!"
+    assert args.project_path.is_dir() and (args.project_path / 'project.json').is_file(), "Invalid Project path provided!"
+
+    #commands are based on 
+    #https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md
+
+    
+    #Build o3de-multiplayersample and the engine (non-monolithic)
+    if args.build_non_mono_tools:
+        process_command(['cmake', '-S', '.', '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
+
+        process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'Editor', 'AssetBundler', 'AssetBundlerBatch', 'AssetProcessor', 'AssetProcessorBatch', '--config','profile'], cwd=args.engine_path)
+
+        asset_processor_batch_path = non_mono_build_path / 'bin' / 'profile' / 'AssetProcessorBatch'
+        process_command([asset_processor_batch_path, '--project-path', args.project_path], cwd=args.engine_path)
+    
+    #Build monolithic game
+    process_command(['cmake', '-S', '.', '-B', str(mono_build_path), '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
+    
+    process_command(['cmake', '--build', str(mono_build_path), '--target', 'MultiplayerSample.GameLauncher', 'MultiplayerSample.ServerLauncher', 'MultiplayerSample.UnifiedLauncher', '--config', args.config], cwd=args.engine_path)
+
+    #Bundle content
+    asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / 'AssetBundlerBatch'
+    engine_asset_list_path = args.project_path / 'AssetBundling' /  'AssetLists' / 'engine_pc.assetlist'
+    
+    process_command([asset_bundler_batch_path, 'assetLists','--addDefaultSeedListFiles', '--assetListFile', engine_asset_list_path, '--project-path', args.project_path, '--allowOverwrites' ], cwd=args.engine_path)
+
+
+    game_asset_list_path = args.project_path /'AssetBundling'/'AssetLists'/'game_pc.assetlist'
+    seed_folder_path = args.project_path/'AssetBundling'/'SeedLists'
+
+    game_asset_list_command = [asset_bundler_batch_path, 'assetLists', '--assetListFile', game_asset_list_path, 
+                    '--seedListFile', seed_folder_path  / 'BasePopcornFxSeedList.seed',
+                    '--seedListFile', seed_folder_path  / 'GameSeedList.seed']
+
+    if args.config == 'profile':
+        game_asset_list_command += ['--seedListFile', seed_folder_path / 'ProfileOnlySeedList.seed']
+
+    game_asset_list_command += ['--seedListFile', seed_folder_path / 'VFXSeedList.seed', '--project-path', args.project_path, '--allowOverwrites']
+
+    process_command(game_asset_list_command, cwd=args.engine_path)
+
+    engine_bundle_path = args.project_path / 'AssetBundling' / 'Bundles' / 'engine_pc.pak'
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', engine_asset_list_path, '--outputBundlePath', engine_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+    # This is to prevent any accidental file locking mechanism from failing subsequent bundling operations
+    time.sleep(1)
+
+    game_bundle_path   = args.project_path / 'AssetBundling' / 'Bundles' / 'game_pc.pak'
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', game_asset_list_path, '--outputBundlePath', game_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+
+    #Create Launcher Zip File
+    os.makedirs( args.output_path / 'Cache' / 'pc', exist_ok=True)
+    os.makedirs( args.output_path / 'Gems' / 'AWSCore', exist_ok=True)
+
+    process_command(['xcopy', '/Y', pathlib.PurePath(args.project_path / 'AssetBundling' / 'Bundles' / '*.pak'),
+                            args.output_path / 'Cache' / 'pc'], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(mono_build_path / 'bin' / 'profile' / '*.*'),
+                            args.output_path], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(mono_build_path  / 'bin' / 'profile' / 'Gems' / 'AWSCore' / '*.*'),
+                            args.output_path / 'Gems' / 'AWSCore'], cwd=args.engine_path)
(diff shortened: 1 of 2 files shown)


ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+from queue import Queue, Empty
+from threading  import Thread
+from typing import List
+from subprocess import Popen, PIPE
+
+logger = logging.getLogger('o3de.gamejam')
+LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
+#This is an export script for MPS on the Windows platform
```
AMZN-Gene, line 25: Space after #

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', engine_asset_list_path, '--outputBundlePath', engine_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+    # This is to prevent any accidental file locking mechanism from failing subsequent bundling operations
+    time.sleep(1)
+
+    game_bundle_path   = args.project_path / 'AssetBundling' / 'Bundles' / 'game_pc.pak'
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', game_asset_list_path, '--outputBundlePath', game_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+
+    #Create Launcher Zip File
```
AMZN-Gene, line 265: Space after #

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+from typing import List
+from subprocess import Popen, PIPE
+
+logger = logging.getLogger('o3de.gamejam')
+LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
+#This is an export script for MPS on the Windows platform
+# this has to be a complete standalone script, b/c project export doesnt exist in main branch yet
+
+#View the argparse parameters for options available. An example invocation:
```
AMZN-Gene, line 28: Space after #

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    assert args.engine_path.is_dir() and (args.engine_path / 'engine.json').is_file(), "Invalid Engine path provided!"
+    assert args.project_path.is_dir() and (args.project_path / 'project.json').is_file(), "Invalid Project path provided!"
+
+    #commands are based on 
+    #https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md
+
+    
+    #Build o3de-multiplayersample and the engine (non-monolithic)
+    if args.build_non_mono_tools:
+        process_command(['cmake', '-S', '.', '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
```
AMZN-Gene, line 222: I might have missed it, but doesn't look like there's any cmake -G, so it's required users have defined a generator before running this. I guess default is used? Not sure. Is there a way to check for that and log an error with steps to fix for the user?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+        return 1
+    return CLICommand(args, cwd, logging.getLogger(), env=env).run()
+
+
+# EXPORT SCRIPT STARTS HERE!
+
+
+if __name__ == "__main__":
+    
+
```
AMZN-Gene, line 191: Is it common to add newlines after __main__?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+from queue import Queue, Empty
+from threading  import Thread
+from typing import List
+from subprocess import Popen, PIPE
+
+logger = logging.getLogger('o3de.gamejam')
+LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
+logging.basicConfig(format=LOG_FORMAT)
+#This is an export script for MPS on the Windows platform
+# this has to be a complete standalone script, b/c project export doesnt exist in main branch yet
```
AMZN-Gene, line 26: This PR is going into development, not main. Did you mean for that?

AMZN-Gene (approved):

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    os.makedirs( args.output_path / 'Cache' / 'pc', exist_ok=True)
+    os.makedirs( args.output_path / 'Gems' / 'AWSCore', exist_ok=True)
+
+    process_command(['xcopy', '/Y', pathlib.PurePath(args.project_path / 'AssetBundling' / 'Bundles' / '*.pak'),
+                            args.output_path / 'Cache' / 'pc'], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(mono_build_path / 'bin' / 'profile' / '*.*'),
+                            args.output_path], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(mono_build_path  / 'bin' / 'profile' / 'Gems' / 'AWSCore' / '*.*'),
+                            args.output_path / 'Gems' / 'AWSCore'], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(args.project_path / 'launch_*.*'), args.output_path], cwd=args.engine_path)
```
lumberyard-employee-dm, line 275: Did you see this feedback about using the [cmake --copy-if-different](https://github.com/o3de/o3de/pull/15955#discussion_r1200655183) command?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+    if args.zip_output:
+        import shutil
+        archive_name = args.output_path
+        print("Zipping output directory (this may take a while)...")
```
lumberyard-employee-dm, line 280: Should these print statements be sent to the logger at INFO level?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+        shutil.make_archive(args.output_path, 'zip', root_dir = args.output_path)
+
+    print(f"Exporting project is complete! Release Directory can be found at {args.output_path}")
```
lumberyard-employee-dm, line 283: repeat: The print statements should be sent to logger.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    parser.add_argument('-egn', '--engine-path', type=pathlib.Path, required=True, help='Path to the intended O3DE engine copy.')
+    parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+    parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'], help='The CMake build configuration to use when building project binaries.')
+    parser.add_argument('-ll', '--log-level', default='INFO',
+                        choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
+                        help="Set the log level")
+    parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
+    parser.add_argument('-nmbp', '--non-mono-build-path', type=pathlib.Path, default=None)
+    parser.add_argument('-mbp', '--mono-build-path', type=pathlib.Path, default=None)
+    parser.add_argument('-zip', '--zip-output', action='store_true', help='This option places the final output of the build into a zipped archive')
```
lumberyard-employee-dm, line 204: A [quiet mode](https://unix.stackexchange.com/questions/191254/why-do-many-commands-provide-a-quiet-option) parameter (`-q/--quiet`) helps with script-ability.
Being able to run this script with no console output would be useful for running the script as part of a chain of other commands

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+                            args.output_path], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(mono_build_path  / 'bin' / 'profile' / 'Gems' / 'AWSCore' / '*.*'),
+                            args.output_path / 'Gems' / 'AWSCore'], cwd=args.engine_path)
+    process_command(['xcopy', '/Y', pathlib.PurePath(args.project_path / 'launch_*.*'), args.output_path], cwd=args.engine_path)
+
+    if args.zip_output:
+        import shutil
+        archive_name = args.output_path
+        print("Zipping output directory (this may take a while)...")
+        shutil.make_archive(args.output_path, 'zip', root_dir = args.output_path)
```
lumberyard-employee-dm, line 281: Why not make the 'zip' option an argparse option?
That way the user can choose whether they want to create a zip or a tar.xz or a tar.gz or  tar.bz

The [shutil.make_archive](https://docs.python.org/3/library/shutil.html#shutil.make_archive) function supports all those formats.
For example
```python
    parser.add_argument(
        "--archive-output-format", "-aof"
        type=str,
        help="Format of archive to outptu",
        choices=["none", "zip", "gzip", "bz2", "xz"],
        default="none"
    )
```

ExportScripts/export_standalone_monolithic_windows.py
lumberyard-employee-dm: No cmake will use the a default generator based on what the user has installed.

On Windows the order would be 
1. Latest version of Visual Studio installed
2. Nmake

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+if __name__ == "__main__":
+    
+
+    parser = argparse.ArgumentParser(prog='Exporter for MultiplayerSample on windows',
+                                 description = "Exports O3DE's MultiplayerSample to the desired install directory in project layout...")
+    parser.add_argument('-ps', '--project-path', type=pathlib.Path, required=True, help='Path to the intended O3DE project.')
+    parser.add_argument('-egn', '--engine-path', type=pathlib.Path, required=True, help='Path to the intended O3DE engine copy.')
+    parser.add_argument('-out', '--output-path', type=pathlib.Path, required=True, help='Path that describes the final resulting Release Directory path location.')
+    parser.add_argument('-cfg', '--config', type=str, default='profile', choices=['release', 'profile'], help='The CMake build configuration to use when building project binaries.')
```
lumberyard-employee-dm, line 197: Can you add an additional statement to the help message to indicate that when tool binaries are built they would be done in profile configuration?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    args = parser.parse_args()
+
+    logging.getLogger().setLevel(args.log_level)
+
+    non_mono_build_path = (args.engine_path) / 'build' / 'non_mono' if args.non_mono_build_path is None else args.non_mono_build_path
+    mono_build_path = (args.engine_path) / 'build' / 'mono' if args.mono_build_path is None else args.mono_build_path
+
+    #validation
+    assert args.engine_path.is_dir() and (args.engine_path / 'engine.json').is_file(), "Invalid Engine path provided!"
+    assert args.project_path.is_dir() and (args.project_path / 'project.json').is_file(), "Invalid Project path provided!"
```
lumberyard-employee-dm, line 214: This shouldf call the o3de CLI utils validation functions for the engine.json and project.json respectively
https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/validation.py#L40
https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/validation.py#L67

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+    if args.zip_output:
+        import shutil
+        archive_name = args.output_path
+        print("Zipping output directory (this may take a while)...")
+        shutil.make_archive(args.output_path, 'zip', root_dir = args.output_path)
+
+    print(f"Exporting project is complete! Release Directory can be found at {args.output_path}")
+
+    process_command(['explorer', args.output_path])
```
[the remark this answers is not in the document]
lumberyard-employee-dm: I recommend having the `quiet` option skip such an operation.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+    game_asset_list_command += ['--seedListFile', seed_folder_path / 'VFXSeedList.seed', '--project-path', args.project_path, '--allowOverwrites']
+
+    process_command(game_asset_list_command, cwd=args.engine_path)
+
+    engine_bundle_path = args.project_path / 'AssetBundling' / 'Bundles' / 'engine_pc.pak'
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', engine_asset_list_path, '--outputBundlePath', engine_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+    # This is to prevent any accidental file locking mechanism from failing subsequent bundling operations
+    time.sleep(1)
```
lumberyard-employee-dm, line 259: A sleep call never fixes any kind of timing issues.

The issue will just manifest itself again.
It sounds like the real fix is to have the SystemFile_WinAPI.cpp [PlatformClose()](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Common/WinAPI/AzCore/IO/SystemFile_WinAPI.cpp#L159-L165) function to invoke the [Flush()](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/Platform/Common/WinAPI/AzCore/IO/SystemFile_WinAPI.cpp#L318-L324) before closing the handle.
That will make sure that the bundle files have been written to disk completely.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+    game_bundle_path   = args.project_path / 'AssetBundling' / 'Bundles' / 'game_pc.pak'
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', game_asset_list_path, '--outputBundlePath', game_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+
+    #Create Launcher Zip File
+    os.makedirs( args.output_path / 'Cache' / 'pc', exist_ok=True)
+    os.makedirs( args.output_path / 'Gems' / 'AWSCore', exist_ok=True)
```
lumberyard-employee-dm, line 267: space after `(`
```suggestion
    os.makedirs(args.output_path / 'Cache' / 'pc', exist_ok=True)
    os.makedirs(args.output_path / 'Gems' / 'AWSCore', exist_ok=True)
```

ExportScripts/export_standalone_monolithic_windows.py
lumberyard-employee-dm: Also should the comment be about creating the launcher layout directory?

What is the fascination for zipping a directory?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+    #Build monolithic game
+    process_command(['cmake', '-S', '.', '-B', str(mono_build_path), '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
+    
+    process_command(['cmake', '--build', str(mono_build_path), '--target', 'MultiplayerSample.GameLauncher', 'MultiplayerSample.ServerLauncher', 'MultiplayerSample.UnifiedLauncher', '--config', args.config], cwd=args.engine_path)
+
+    #Bundle content
+    asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / 'AssetBundlerBatch'
+    engine_asset_list_path = args.project_path / 'AssetBundling' /  'AssetLists' / 'engine_pc.assetlist'
+    
+    process_command([asset_bundler_batch_path, 'assetLists','--addDefaultSeedListFiles', '--assetListFile', engine_asset_list_path, '--project-path', args.project_path, '--allowOverwrites' ], cwd=args.engine_path)
```
lumberyard-employee-dm, line 238: Before the AssetBundlerBatch runs, the AssetProcessorBatch needs to be run to make sure the assets are processed.

lumberyard-employee-dm (changes requested):

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+    #commands are based on 
+    #https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md
+
+    
+    #Build o3de-multiplayersample and the engine (non-monolithic)
+    if args.build_non_mono_tools:
+        process_command(['cmake', '-S', '.', '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
+
+        process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'Editor', 'AssetBundler', 'AssetBundlerBatch', 'AssetProcessor', 'AssetProcessorBatch', '--config','profile'], cwd=args.engine_path)
```
lumberyard-employee-dm, line 224: Build time can be saved by skipping building the Editor, since it is not needed for the export process

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,285 @@
...
+
+        asset_processor_batch_path = non_mono_build_path / 'bin' / 'profile' / 'AssetProcessorBatch'
+        process_command([asset_processor_batch_path, '--project-path', args.project_path], cwd=args.engine_path)
```
AMZN-Olex, line 227: Another option is to build MultiplayerSample.Assets target with cmake which should build the assets for you.

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: That is worth trying, but one question I have is if we anticipate this python script being used outside of our python install (say the user's python install). If that is the case, then we can't guarantee that the o3de CLI is available as a site-package (unless we get it installed in pypi).

lumberyard-employee-dm: It would be impossible to use the this python script outside of the o3de CLI because the export-project script itself is part of [o3de CLI](https://github.com/o3de/o3de/blob/development/scripts/o3de/o3de/export_project.py)

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: I can reword the comment here to be about launcher layout directory. Originally I copied headings from Packed Asset Builds 1:1, which annotated this section like the original comment.

Right now the actual zipping operation is an optional operation, gated by the `--zip-output` parameter. If ommitted, the default value is false.

ExportScripts/export_standalone_monolithic_windows.py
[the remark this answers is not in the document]
tkothadev: Yes the intention was to replicate Windows unzip by showing the result of the build process. But based on the instructions for PackedAssetBuilds, that's not as helpful as I originally thought, because you have to run the executables via command line (if you can double click the exe file, then it's still helpful I think). Maybe a new terminal window at the output location would be easier to work with.

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: excellent point, will go ahead and use your suggestion then

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: Do we want errors to still be emitted? If so I can have the quiet flag to indicate "only emit on error"

lumberyard-employee-dm: Yes the quiet parameter should still output errors.
It is just informational and trace messages should be supressed

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: For now yes. I believe the O3DE contribution protocol calls for only commits processed in stabilization to go to main, which I would want at least this initial standalone script to eventually go. However currently project export does not exist in the main branch, so this is the closest compromise I could think of. Prepare and submit a standalone script, so that it can at lease be manually copied over.

lumberyard-employee-dm: Main is an already a snapshot of the latest released version.
Since this change is after the release, it should go into the development and come out with the next release.

There is no need to try to backport functionality to an older release, plus it would require a stabilization period and support from the SIG-release group to try to create a point release(i.e 2205.1)

ExportScripts/export_standalone_monolithic_windows.py
lumberyard-employee-dm: Yes can you add a GHI for this issue.

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: Yes I did. Per the documentation you linked, it looks like CMake copy commands do not include wildcard patterns, so I would need to copy individual folders or file items.

I'm thinking one alternative is to use `shutil` with glob patterns.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -258,28 +262,33 @@ def process_command(args: list,
...
-        shutil.make_archive(args.output_path, 'zip', root_dir = args.output_path)
+        logger.info("Archiving output directory (this may take a while)...")
+        shutil.make_archive(args.output_path, args.archive_output_format, root_dir = args.output_path)
 
-    print(f"Exporting project is complete! Release Directory can be found at {args.output_path}")
+    logger.info(f"Exporting project is complete! Release Directory can be found at {args.output_path}")
 
-    process_command(['explorer', args.output_path])
+    if args.quiet:
+        process_command(['explorer', args.output_path])
```
lumberyard-employee-dm, line 294: This check should be inverted to check if its not quiet
```suggestion
    if not args.quiet:
        process_command(['explorer', args.output_path])
```

lumberyard-employee-dm (approved):

AMZN-Olex (approved):
Now onto testing it in development branches.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+import time
+import glob
+
+from o3de.validation import valid_o3de_project_json, valid_o3de_engine_json
+from queue import Queue, Empty
+from threading  import Thread
+from typing import List
+from subprocess import Popen, PIPE
+
+logger = logging.getLogger('o3de.gamejam')
```
AMZN-alexpete, line 21: `o3de.gamejam` ? I think something like  `MultiplayerSample Export` might make more sense

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+# EXPORT SCRIPT STARTS HERE!
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(prog='Exporter for MultiplayerSample on windows',
+                                 description = "Exports O3DE's MultiplayerSample to the desired install directory in project layout...")
+    parser.add_argument('-ps', '--project-path', type=pathlib.Path, required=True, help='Path to the intended O3DE project.')
+    parser.add_argument('-egn', '--engine-path', type=pathlib.Path, required=True, help='Path to the intended O3DE engine copy.')
```
AMZN-alexpete, line 192: what do you mean by `engine copy`,  does this just mean path to the engine?

AMZN-alexpete: fwiw I would have recommended you stick with the short abbreviations that match the first letter of the long argument names.  e.g. '--project-path' '-pp' ,  '--engine-path', -'ep' that's also what we typically do for `o3de.py`

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: Missed that one! Thanks for the catch, that'll be updated in a future PR

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: Yes, that should mean path to the engine (as in 'path to the copy of the O3DE engine residing on your hard drive').

Regarding abbreviations, I think you're right, and that'll be updated in another PR.

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: Issue made here: https://github.com/o3de/o3de/issues/16167

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+    assert valid_o3de_project_json(args.project_path / 'project.json') and valid_o3de_engine_json(args.engine_path / 'engine.json')
+
+
+    #commands are based on 
+    #https://github.com/o3de/o3de-multiplayersample/blob/development/Documentation/PackedAssetBuilds.md
+
+    
+    #Build o3de-multiplayersample and the engine (non-monolithic)
+    if args.build_non_mono_tools:
+        process_command(['cmake', '-S', '.', '-B', str(non_mono_build_path), '-DLY_MONOLITHIC_GAME=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
```
AMZN-alexpete, line 227: will this work if the project path has spaces in it?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+
+        process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'AssetBundler', 'AssetBundlerBatch', 'AssetProcessor', 'AssetProcessorBatch', '--config','profile'], cwd=args.engine_path)
+
+        process_command(['cmake', '--build', str(non_mono_build_path), '--target', 'MultiplayerSample.Assets', '--config', 'profile'], cwd=args.engine_path)
+    
+    #Build monolithic game
+    process_command(['cmake', '-S', '.', '-B', str(mono_build_path), '-DLY_MONOLITHIC_GAME=1', '-DALLOW_SETTINGS_REGISTRY_DEVELOPMENT_OVERRIDES=0', f'-DLY_PROJECTS={args.project_path}'], cwd=args.engine_path)
```
AMZN-alexpete, line 234: does this work if the project path has spaces in it?

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+                    '--seedListFile', seed_folder_path  / 'GameSeedList.seed']
+
+    if args.config == 'profile':
+        game_asset_list_command += ['--seedListFile', seed_folder_path / 'ProfileOnlySeedList.seed']
+
+    game_asset_list_command += ['--seedListFile', seed_folder_path / 'VFXSeedList.seed', '--project-path', args.project_path, '--allowOverwrites']
+
+    process_command(game_asset_list_command, cwd=args.engine_path)
+
+    engine_bundle_path = args.project_path / 'AssetBundling' / 'Bundles' / 'engine_pc.pak'
```
AMZN-alexpete, line 259: I'd recommend writing directly into the release output path instead of into the project folder which is likely under source control.

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -0,0 +1,294 @@
...
+    process_command([asset_bundler_batch_path, 'bundles', '--assetListFile', engine_asset_list_path, '--outputBundlePath', engine_bundle_path, '--project-path', args.project_path, '--allowOverwrites'], cwd=args.engine_path)
+
+    # This is to prevent any accidental file locking mechanism from failing subsequent bundling operations
+    time.sleep(1)
+
+    game_bundle_path = args.project_path / 'AssetBundling' / 'Bundles' / 'game_pc.pak'
```
AMZN-alexpete, line 265: ditto here, would recommend writing directly to release output folder and not into project folder

ExportScripts/export_standalone_monolithic_windows.py
lumberyard-employee-dm: It should because the arguments are passed via list argument and not parsed from a string.
So the application receives the entire path as a single argv[n] entry.

PR #442, opened 2023-06-22
Support MultiProcess GameLift Server Logs

AMZN-Gene:
- Updating MPSGameLift gem so that server logs are saved out to a location based on the time and server pid (process id). This is useful for GameLift which will run multiple servers at once and will save the server.log files if the server crashes. Before this change, the log folder would contain server.log, server(1).log, server(2).log etc making it hard to tell which logs belongs to which server. After this change, there are separate log folders per server (each folder name is the pid).
- Updating GameLift readme to include sv_gameSessionNoPlayerShutdownTimeoutSeconds; using the default value in the readme, but it's good to know it exists.

Tested by running multiplayer servers on GameLift and intentionally terminating a server with a non-0 exit code so that GameLift will grab the logs. The log folder archived only contains the 1 server.log file

Here's the code used to intentionally terminate the server with a non-zero exit code.
`AzFramework::ApplicationRequests::Bus::Broadcast(&AzFramework::ApplicationRequests::TerminateOnError, 42);`


MPSGameLift/Code/Source/MPSGameLiftModuleInterface.h
```diff
@@ -33,6 +36,20 @@ namespace MPSGameLift
...
+                // This way GameLift can archive the logs for a specific server.
+                AZ::IO::FixedMaxPath projectLogPath;
+                AZStd::fixed_string<32> processIdString;
+
+                AZ::SettingsRegistry::Get()->Get(projectLogPath.Native(), AZ::SettingsRegistryMergeUtils::FilePathKey_ProjectLogPath);
+                AZStd::to_string(processIdString, AZ::Platform::GetCurrentProcessId());
+                projectLogPath = projectLogPath / processIdString;
+
+                AZ::SettingsRegistry::Get()->Set(AZ::SettingsRegistryMergeUtils::FilePathKey_ProjectLogPath, projectLogPath.Native());
+            #endif
```
lumberyard-employee-dm, line 51: Luckily this is restricted to the MultiplayerSample project because otherwise the MPSGameLift gem is overstepping its bounds by changing the LogPath of ALL log files.

AMZN-Olex (approved):

MPSGameLift/Code/Source/MPSGameLiftModuleInterface.h
AMZN-Gene: I think any server using GameLift would want to move their server logs to a new folder.

lumberyard-employee-dm: That is presumptuous of GameLift redirecting all logs to a different folder including non-GameLift logs
image

I would understand if GameLift only redirected the Server.log files to a different location, but not modify the application wide log directory.

I could understand if the [LogFileBaseName](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Logging/LoggingComponent.h#L19-L50) in the LoggingComponent was updated to prepend the process ID as a folder to the location where to write the logs to in this logic.

But overriding the entire Log Path can be seen as devious behavior by other Gem modules and it could make it harder to troubleshoot an issue where the log files end up in a location that is unexpected.
It wouldn't be documented why the logging directory changed on the user from normal application.

AMZN-Gene: I don't think there are "GameLift" specific logs. This change is so GameLift knows which logs belong to which server. Wouldn't GameLift want all those files? Anything in there could be valuable information about why the server crashed.

lumberyard-employee-dm: Correction: It is actually the [SystemInit parameters](https://github.com/o3de/o3de/blob/development/Code/LauncherUnified/Launcher.cpp#L499) that sets the Log file path through CrySystem.

For the Server that value is [@log@/Server.log](https://github.com/o3de/o3de/blob/development/Code/LauncherUnified/Server.cpp#L22-L25).

That is finally set in the Cry LogFile as part [CSystem::Init](https://github.com/o3de/o3de/blob/development/Code/Legacy/CrySystem/SystemInit.cpp#L872-L876)

lumberyard-employee-dm: I am saying that the responsibility of setting the log directory is outside of the GameLift "pay grade".
If GameLift is allowed to do that, why can't the Atom do the same and override the log directory to "<render-api>/<process-id>"?

AMZN-Gene: Ah, I see. Would you rather it be a cvar? IE: make it more explicit that the user has to actively enable the log change?

lumberyard-employee-dm: Well this change is fine, because this Gem is specific to the "MultiplayerSample" project and therefore it can be treated as the "MultiplayerSample" is changing the log directory.

Pretty much we don't want non-project related Gems to affect shared project resources, like the Log directory or the User directory location.

AMZN-Gene: I want a generic solution. Anyone wanting to run GameLift could drop this gem into their project, so it would be cool to convert this gem into a generic (non-project specific) gem soon.

lumberyard-employee-dm: Well rembmer process IDs can be recycled.
So there is the possibility that the same process ID appears if the ServerLauncher is launched multiple times.

lumberyard-employee-dm: > I want a generic solution. Anyone wanting to run GameLift could drop this gem into their project, so it would be cool to convert this gem into a generic (non-project specific) gem soon.

Yeah, this seems more like a general problem that shouldn't be just restricted to the ServerLauncher.

I imagine all of the Applications like the Editor, AssetProcessor, GameLauncher would want some kind of way to segment logs into a different structure.
We could even add meta File IO aliases such as `@processid@` or `@timenow@` that can be used to substitute those alias placeholders with the ProcessID or the current Time.
Imagine, the project could then specify the path to the log settings via a `@log@/@processid@/Server.log

AMZN-Gene: > Well remember process IDs can be recycled. So there is the possibility that the same process ID appears if the ServerLauncher is launched multiple times.
> 
> For the Core event logger we use a combination of a ISO8601 Timestamp and Process ID to try to make sure that filenames are unique: https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/ComponentApplication.cpp#L670-L681

Nice! Upgrade so the folder is _user/log/<timestamp>_<process-id>/_

lumberyard-employee-dm (approved):

Issue #438, opened 2023-06-15
Update Export_standalone_monolithic_windows.py to Validate Necessary Engine Exes Exists


AMZN-Gene:
Running Export_standalone_monolithic_windows.py will end in a python exception if using a project-centric (non-sdk) build of Multiplayer Sample

**Describe the solution you'd like**
Update Export_standalone_monolithic_windows.py to first check if AssetBundlerBatch exists
See line #424

Log an error/warning explaining to the user that the exe is missing and a steps to correct the problem (either switching to engine centric, or keeping project centric and building by hand).

PR #443, opened 2023-06-22
Mps export feedback patch

tkothadev:
This PR is for addressing feedback left at the end of the previous PR: https://github.com/o3de/o3de-multiplayersample/pull/433

This PR also includes a fix for the following issue: https://github.com/o3de/o3de-multiplayersample/issues/438

#### Testing
The export script has now been tested with the development branch and main branch of o3de and o3de-multiplayer-sample on windows.

The script is able to create a release bundle ready for play using both the release and profile configurations.

#### Out of Scope
This PR does not address adding a testing/validation layer for the script, nor the ability to swap to project-centric  mode. That will be included in a future PR. This is just to fix issues with the existing script.
This script is also only tested for windows. There will be a dedicated PR for updating the script to be cross platform for Linux.

### DIFF
--- a/ExportScripts/export_standalone_monolithic_windows.py
+++ b/ExportScripts/export_standalone_monolithic_windows.py
@@ -18,7 +18,7 @@ from threading  import Thread
 from typing import List
 from subprocess import Popen, PIPE
 
-logger = logging.getLogger('o3de.gamejam')
+logger = logging.getLogger('o3de.mps_export')
 LOG_FORMAT = '[%(levelname)s] %(name)s: %(message)s'
 logging.basicConfig(format=LOG_FORMAT)
 # This is an export script for MPS on the Windows platform
@@ -27,7 +27,7 @@ logging.basicConfig(format=LOG_FORMAT)
 # View the argparse parameters for options available. An example invocation:
 
 # @<O3DE_ENGINE_ROOT_PATH>
-# > python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -ps <O3DE_MPS_PROJECT_ROOT_PATH> -egn <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -zip
+# > python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -pp <O3DE_MPS_PROJECT_ROOT_PATH> -ep <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -a -aof zip
 
 def enqueue_output(out, queue):
     for line in iter(out.readline, b''):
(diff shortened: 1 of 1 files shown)


ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -235,8 +248,18 @@ def process_command(args: list,
     
     process_command(['cmake', '--build', str(mono_build_path), '--target', 'MultiplayerSample.GameLauncher', 'MultiplayerSample.ServerLauncher', 'MultiplayerSample.UnifiedLauncher', '--config', args.config], cwd=args.engine_path)
 
-    #Bundle content
+    #Before bundling content, make sure that the necessary executables exist
     asset_bundler_batch_path = non_mono_build_path / 'bin' / 'profile' / 'AssetBundlerBatch'
+    if not asset_bundler_batch_path.is_file():
+        logger.error(f"AssetBundlerBatch not found at path '{asset_bundler_batch_path}'. In order to bundle the data for MPS, this executable must be present!")
+        logger.error("To correct this issue, do 1 of the following: "
+                     "1) use the --build-non-mono-tools flag in the CLI parameters"
```
AMZN-Gene, line 256: Capital "Use"

AMZN-Gene (approved):

AMZN-alexpete (commented):
What happens if you provide `--archive-output` but don't specify `--archive-output-format`? it looks like the format will default to none and pass that to `shutil.make_archive` but I can't tell from the docs whether that's OK or not

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -27,7 +27,7 @@
 # View the argparse parameters for options available. An example invocation:
 
 # @<O3DE_ENGINE_ROOT_PATH>
-# > python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -ps <O3DE_MPS_PROJECT_ROOT_PATH> -egn <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -zip
+# > python\python.cmd <O3DE_MPS_PROJECT_ROOT_PATH>\ExportScripts\export_standalone_monolithic_windows.py -pp <O3DE_MPS_PROJECT_ROOT_PATH> -ep <O3DE_ENGINE_ROOT_PATH> -bnmt -out <MPS_OUTPUT_RELEASE_DIR_PATH> -a -aof zip
```
AMZN-alexpete, line 30: it seems like we could just have a single parameter e.g. `--archive-output-format`, and if it isn't set then we don't make an archive rather than requiring 2 parameters - unless we change how `archive-output-format` works to where it defaults to something

AMZN-alexpete: It looks like if you don't provide a default value for the archive format the script will error out:
```
>>> shutil.make_archive('f:/Amazon/test', None, 'f:/Amazon/Lumberyard_1.27')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python38\lib\shutil.py", line 1034, in make_archive
    raise ValueError("unknown archive format '%s'" % format) from None
ValueError: unknown archive format 'None'
```

tkothadev: updating it to set the default to zip if not specified.

AMZN-alexpete (approved):

ExportScripts/export_standalone_monolithic_windows.py
```diff
@@ -183,22 +183,24 @@ def process_command(args: list,
...
+                        help='The CMake build configuration to use when building project binaries. If tool binaries are built with this script, they will use profile mode.')
     parser.add_argument('-ll', '--log-level', default='ERROR',
                         choices=['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'],
                         help="Set the log level")
     parser.add_argument('-aof', '--archive-output-format',
                         type=str,
                         help="Format of archive to create from the output directory",
-                        choices=["none", "zip", "gzip", "bz2", "xz"], default="none")
+                        choices=["zip", "gzip", "bz2", "xz"], default="zip")
     parser.add_argument('-bnmt', '--build-non-mono-tools', action='store_true')
```
AMZN-Olex, line 204: Is this on by default now or off?

AMZN-Olex (approved):

ExportScripts/export_standalone_monolithic_windows.py
tkothadev: it is still off by default, as there is a separate flag to decide whether to archive or not (`--archive-output`)

AMZN-Olex: I mean specifically: '--build-non-mono-tools'

PR #447, opened 2023-07-13
MPSGameLift Gem: AWS Region Latency Finder 

AMZN-Gene:
Adding a system component to retrieve the latency from various AWS regions.
By getting the round-trip-time to different AWS endpoints (us-west-2, us-east-1, etc) players can select the best server for their matches.

A digestible part of a large initiative for integrating GameLift's FlexMatch. 
(https://github.com/aws-lumberyard-dev/o3de-multiplayersample/tree/mps_flex_match)

Tested by creating a menu component and calling "RequestLatencies". Ensured OnRequestLatenciesComplete was received, and that latency times are correct.

### DIFF
--- a/MPSGameLift/Code/CMakeLists.txt
+++ b/MPSGameLift/Code/CMakeLists.txt
@@ -59,14 +59,12 @@ ly_add_target(
             Gem::HttpRequestor
             Gem::Multiplayer.Client
             Gem::LyShine
-
         PRIVATE
             Gem::Multiplayer.Client.Static
             Gem::AWSCore.Static
             Gem::HttpRequestor.Static
             Gem::AWSGameLift.Client.Static
             Gem::LyShine.Static
-            3rdParty::AWSNativeSDK::Core
 )
 
 ly_add_target(
--- a/MPSGameLift/Code/CMakeLists.txt
+++ b/MPSGameLift/Code/CMakeLists.txt
@@ -56,14 +56,17 @@ ly_add_target(
         PUBLIC
             AZ::AzCore
             AZ::AzFramework
+            Gem::HttpRequestor
             Gem::Multiplayer.Client
             Gem::LyShine
 
         PRIVATE
             Gem::Multiplayer.Client.Static
             Gem::AWSCore.Static
+            Gem::HttpRequestor.Static
             Gem::AWSGameLift.Client.Static
             Gem::LyShine.Static
+            3rdParty::AWSNativeSDK::Core
 )
 
 ly_add_target(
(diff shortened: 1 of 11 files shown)


MPSGameLift/Code/CMakeLists.txt
```diff
@@ -56,14 +56,17 @@ ly_add_target(
...
             Gem::Multiplayer.Client
             Gem::LyShine
 
         PRIVATE
             Gem::Multiplayer.Client.Static
             Gem::AWSCore.Static
+            Gem::HttpRequestor.Static
             Gem::AWSGameLift.Client.Static
             Gem::LyShine.Static
+            3rdParty::AWSNativeSDK::Core
```
AMZN-alexpete, line 69: Are you sure this is needed? AWSCore has this already in PUBLIC https://github.com/o3de/o3de/blob/development/Gems/AWSCore/Code/CMakeLists.txt#L34

MPSGameLift/Code/CMakeLists.txt
```diff
@@ -234,6 +242,8 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
                 Gem::Multiplayer.Unified.Static
                 Gem::AWSCore.Static
                 Gem::AWSGameLift.Client.Static
+                Gem::HttpRequestor.Static
+                3rdParty::AWSNativeSDK::Core
```
AMZN-alexpete, line 246: Not sure if this is needed because it already depends on AWSCore

MPSGameLift/Code/Source/RegionalLatencySystemComponent.cpp
```diff
@@ -0,0 +1,102 @@
...
+                    {
+                        AZ_Error("RegionalLatencySystemComponent", false, "Failed to receive response for region: %s.", region);
+                    }
+
+                    // Check if all requests have responded
+                    m_responsesPending.fetch_sub(1);
+                    if (m_responsesPending.load() == 0)
+                    {
+                        RegionalLatencyFinderNotificationBus::Broadcast(&RegionalLatencyFinderNotifications::OnRequestLatenciesComplete, m_regionalLatencies);
+                    }
```
AMZN-alexpete, line 84: suggestion, rather than using a lock on `m_regionalLatencies` use a temporary array passed as a local into the callback, and then on line 81 after you get all the responses, use AZ::TickBus::Enqueue to enqueue a lambda that sets the m_responsesPending map on the main thread and then does the broadcast after.  That way you won't need the mutex and whoever receives the notification doesn't have to worry about receiving a notification from thread that isn't the main thread.

AMZN-alexpete (approved):

PR #450, opened 2023-07-21
Client Request Matchmaking

AMZN-Gene:
Client can request matchmaking from a server-less backend and receives a ticket-id response

Note: JSON parser is crashing when hitting arrays or dictionaries if they are ignored, for that reason, player data is being stored even though we only need the ticket-id.

### DIFF
--- a/MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
+++ b/MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
@@ -88,7 +88,6 @@ namespace MPSGameLift
             PlayerAttributes playerAttributes;
         };
 
-
         //! Struct for storing the success response.
         struct RequestMatchmakingResponse
         {
@@ -161,7 +160,7 @@ namespace MPSGameLift
             Parameters parameters; //! Request parameter.
         };
 
-        using MPSRequestMatchmakingRequestJob = AWSCore::ServiceRequestJob<RequestMatchmaking>;
+        using MPSRequestMatchmakingJob = AWSCore::ServiceRequestJob<RequestMatchmaking>;
     }  // ServiceAPI
 
     void MatchmakingSystemComponent::Activate()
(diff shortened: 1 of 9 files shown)


MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+                if (strcmp(key, "type") == 0)
+                {
+                    return reader.Accept(type);
+                }
+
+                return reader.Ignore();
+            }
+
+            //! Do not rename the following members since they are expected by the AWSCore dependency.
```
tkothadev, line 130: Does AWSCore already have a struct that encapsulates these members? Maybe instead of making a specific RequestMatchmakingError struct, you can reuse that struct?

MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+namespace MPSGameLift
+{
+    namespace ServiceAPI
+    {
+        struct PlayerSkill
+        {
+            bool OnJsonKey(const char* key, AWSCore::JsonReader& reader)
+            {
+                if (strcmp(key, "N") == 0)
```
AMZN-Olex, line 28: General comment for these messages. Do we or could we have some docs on this? Just looking at the comparisons here it's not clear what is going on. Now about adding samples of the incoming or outgoing json messages and what keys and values are being looked for in these structs.

MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
AMZN-Gene: Fixed! Thank you... also updated the matchmaking AWS Lambda to send back an error that works with the built in AWSCore::Error

MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp
AMZN-Gene: Thank you! Added comments and adding doc urls:
https://docs.aws.amazon.com/gamelift/latest/apireference/API_MatchmakingTicket.html
https://docs.aws.amazon.com/gamelift/latest/apireference/API_Player.html

AMZN-Olex (approved):

tkothadev (approved):

PR #458, opened 2023-08-09
FlexMatch Connection Menu

AMZN-Gene:
- Menu allowing players to connect via FlexMatch
- Player can press CONNECT or QUIT
- Simple connection status text added for debug information
- Added menu to the game seed list

image

image

PR #460, opened 2023-08-15
Fix formatting error related to linux integer types

spham-amzn:
Fixes the following Linux compile error:
```
PRIu64/home/o3de/github/o3de-multiplayersample/MPSGameLift/Code/Source/MatchmakingSystemComponent.cpp:267:85: error: format specifies type 'long long' but the argument has type 'std::chrono::duration<long, std::ratio<1, 1000>>::rep' (aka 'long') [-Werror,-Wformat]
            httpLatenciesParam += AZStd::string::format("%s_%lld ", region.c_str(), latencyMs.count());
```


AMZN-Gene:
Also just casting to uint32 here: https://github.com/o3de/o3de-multiplayersample/pull/458

PR #476, opened 2024-02-14
Adding Logo

AMZN-Gene:
Adding logo to fix compiler error


Resources/LegacyLogoLauncher.bmp
AMZN-alexpete, line 1: Can you name this `Splash.bmp`? `LegacyLogoLauncher.bmp` is the old name

AMZN-alexpete: Actually, this is fine if you want to keep it backwards compatible with older engines

AMZN-alexpete (approved):

PR #480, opened 2024-06-05
Delete Old GameLift Export Script

AMZN-Gene:
Use the new, built-in O3DE export script for creating the GameLift packages

Tested by building and connecting to GameLift fleet on Windows and AL2023


tkothadev (approved):
Nice! Always great to prune stale code.

PR #481, opened 2024-06-13
Alternate example of multiplier sample entity filtering component using occlusion culling bus

gadams3:
This is another example of a multiplayer entity filtering component that should prevent network replication for players or objects that cannot be seen by the client, if there is an active occlusion culling system that utilizes the occlusion request bus.

The code compiles in the component can be added from the editor but it is not added or active by default.

### DIFF
--- a/Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
+++ b/Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
@@ -47,8 +47,8 @@ namespace MultiplayerSample
     }
 
     bool OcclusionFilteredEntityComponent::IsEntityFiltered(
-        [[maybe_unused]] AZ::Entity* entity,
-        [[maybe_unused]] Multiplayer::ConstNetworkEntityHandle controllerEntity,
+        AZ::Entity* entity,
+        Multiplayer::ConstNetworkEntityHandle controllerEntity,
         [[maybe_unused]] AzNetworking::ConnectionId connectionId)
     {
         bool result = false;
@@ -60,11 +60,11 @@ namespace MultiplayerSample
         {
             AzFramework::OcclusionRequestBus::Broadcast([&](AzFramework::OcclusionRequestBus::Events* occlusionHandler) {
                 // Check for a preexisting occlusion view or create one if necessary.
-                if (!occlusionHandler->IsOcclusionView(m_occlusionViewName))
+                if (!occlusionHandler->IsOcclusionViewValid(m_occlusionViewName))
                 {
                     occlusionHandler->CreateOcclusionView(m_occlusionViewName);
                 }
-                if (occlusionHandler->IsOcclusionView(m_occlusionViewName))
+                if (occlusionHandler->IsOcclusionViewValid(m_occlusionViewName))
                 {
                     // Perform an occlusion query to determine if the controlled entity can see the filtered entity.
                     const AZStd::vector<bool> visibility = occlusionHandler->GetOcclusionViewEntityToEntityVisibility(
--- a/Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
+++ b/Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
@@ -70,7 +70,7 @@ namespace MultiplayerSample
                     const AZStd::vector<bool> visibility = occlusionHandler->GetOcclusionViewEntityToEntityVisibility(
                         m_occlusionViewName, controllerEntity.GetEntity()->GetId(), AZStd::vector<AZ::EntityId>{ entity->GetId() }
                     );
-                    // If the query succeeded and the entity cannot be seend then filter it out from network replication.
+                    // If the query succeeded and the entity cannot be seen then filter it out from network replication.
                     result = !visibility.empty() && !visibility[0];
                 }
                 });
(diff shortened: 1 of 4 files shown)


Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+        if (entityConnectionId != AzNetworking::InvalidConnectionId)
+        {
+            AzFramework::OcclusionRequestBus::Broadcast([&](AzFramework::OcclusionRequestBus::Events* occlusionHandler) {
+                // Check for a preexisting occlusion view or create one if necessary.
+                if (!occlusionHandler->IsOcclusionView(m_occlusionViewName))
+                {
+                    occlusionHandler->CreateOcclusionView(m_occlusionViewName);
+                }
+                if (occlusionHandler->IsOcclusionView(m_occlusionViewName))
+                {
```
forkercat, line 68: Is this "if" check needed? Or maybe we can use "else" here.

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+                    // Perform an occlusion query to determine if the controlled entity can see the filtered entity.
+                    const AZStd::vector<bool> visibility = occlusionHandler->GetOcclusionViewEntityToEntityVisibility(
+                        m_occlusionViewName, controllerEntity.GetEntity()->GetId(), AZStd::vector<AZ::EntityId>{ entity->GetId() }
+                    );
+                    // If the query succeeded and the entity cannot be seend then filter it out from network replication.
```
forkercat, line 73: nit: seen

forkercat (approved):

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+        // If one is available, use the occlusion culling system to filter out players that cannot be seen by the controlled player.
+        const auto entityNetBindComponent = entity->FindComponent<Multiplayer::NetBindComponent>();
+        const auto entityConnectionId = entityNetBindComponent ? entityNetBindComponent->GetOwningConnectionId() : AzNetworking::InvalidConnectionId;
+        if (entityConnectionId != AzNetworking::InvalidConnectionId)
+        {
+            AzFramework::OcclusionRequestBus::Broadcast([&](AzFramework::OcclusionRequestBus::Events* occlusionHandler) {
+                // Check for a preexisting occlusion view or create one if necessary.
+                if (!occlusionHandler->IsOcclusionView(m_occlusionViewName))
+                {
+                    occlusionHandler->CreateOcclusionView(m_occlusionViewName);
```
moudgils, line 65: Can view creation not succeed? Should we handle for that here?

gadams3: > Can view creation not succeed? Should we handle for that here?

It depends on the implementation of the occlusion system. It's possible that a system might need pre baked assets loaded and ready before it can initialize an occlusion view. Until those assets are ready we can assume everything is visible. The current code is handling it but there could be other come a better patterns for that.

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
gadams3: I think it's fine for these things to happen back to back. no need for the else case.

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    void OcclusionFilteredEntityComponent::Deactivate()
+    {
+        AZ::Interface<IFilterEntityManager>::Unregister(this);
+        AzFramework::OcclusionRequestBus::Broadcast(&AzFramework::OcclusionRequestBus::Events::DestroyOcclusionView, m_occlusionViewName);
+    }
+
+    bool OcclusionFilteredEntityComponent::IsEntityFiltered(
+        [[maybe_unused]] AZ::Entity* entity,
```
spham-amzn, line 50: I don't see a need for `[[maybe_unused]]` for this parameter

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+        [[maybe_unused]] Multiplayer::ConstNetworkEntityHandle controllerEntity,
+        [[maybe_unused]] AzNetworking::ConnectionId connectionId)
```
spham-amzn, line 52: Do we need `[[maybe_unused]]` needed here? I don't see these parameters every being conditionally excluded below.

Gem/Code/Source/Components/OcclusionFilteredEntityComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        bool result = false;
+
+        // If one is available, use the occlusion culling system to filter out players that cannot be seen by the controlled player.
+        const auto entityNetBindComponent = entity->FindComponent<Multiplayer::NetBindComponent>();
+        const auto entityConnectionId = entityNetBindComponent ? entityNetBindComponent->GetOwningConnectionId() : AzNetworking::InvalidConnectionId;
+        if (entityConnectionId != AzNetworking::InvalidConnectionId)
```
spham-amzn, line 59: This looks like it qualifies for either an AZ_Warning or AZ_Assert

gadams3: Which part looks like it should be an assert? It can still be an invalid connection id for other entities. This is how I'm filtering out client or player entities

